PR #1, opened 2017-10-03
Initial automation test framework commit

ashwinimore:
Review requested from
@andlar
@kekey1 
@brianlindsy

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,1 @@
+# chpl-test


.classpath
```diff
@@ -0,0 +1,114 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kekey1: .classpath is an eclipse-specific file and we typically don't check those in. If you look at the .gitignore file in chpl (https://github.com/chpladmin/chpl-api/blob/development/.gitignore) you should see a bunch of Eclipse-specific files listed. You can probably copy that .gitignore here and use it without changing a thing.

.project
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kekey1: Should be ignored

.settings/org.eclipse.core.resources.prefs
```diff
@@ -0,0 +1,6 @@
+eclipse.preferences.version=1
```
kekey1: .gitignore

test-output/old/Default suite/methods-alphabetical.html
```diff
@@ -0,0 +1,6 @@
+<h2>Methods run, sorted chronologically</h2><h3>&gt;&gt; means before, &lt;&lt; means after</h3><p/><br/><em>Default suite</em><p/><small><i>(Hover the method name to see the test class name)</i></small><p/>
```
kekey1: Should the output of the tests be checked in? I'm not 100% sure that's what this is, but if so I think this test-output folder could also be ignored.

kekey1 (changes requested):

.gitignore
```diff
@@ -0,0 +1 @@
+/target/
```
andlar: This will keep new files in target from being stored in git, but there are some .class files that are already in there and should be removed.

Also, the entire /test-ouput/ directory should probably be ignored as well

src/test/java/stepDefinitions/loginSteps.java
```diff
@@ -0,0 +1,137 @@
...
+		driver.findElement(By.xpath("//*[@id='admin']/li/div/form/button[1]")).click();
+	}
+	
+	@Then("^login should be successful$")
+	public void login_should_be_successful() throws Throwable {
+		System.out.println("login successfull");
+	}
+}
+	
+
```
andlar: Definitely don't need all of the whitespace at the bottom of this file

PR #2, opened 2017-10-03
Added changes as suggested and removed ignored files

ashwinimore:
Review request
@andlar 
@kekey1 
@brianlindsy


kekey1 (changes requested):
For some reason the test-output is still in here. Maybe it needs to be in .gitignore without the leading or trailing slash?

.gitignore
```diff
@@ -0,0 +1,28 @@
...
+*~.nib
+local.properties
+.classpath
+.settings/
+.loadpath
+/target/
+.settings/org.eclipse.core.resources.prefs 
+test-output/old/Default suite/methods-alphabetical.html
+/test-output/ 
+
```
andlar: Line 19 is a duplicate, given line 16. Line 20 is a duplicate given line 21. Line 21 should be "test-output/. Line 18 should be "target/

I think all of the files in "target/" are auto-generated, so you should `git rm -r --cached target/*` as well

andlar (changes requested):

andlar (approved):

kekey1 (approved):

brianlindsy (approved):

PR #3, opened 2017-10-06
OCD-1784-AQA:Verify SED text updates

ashwinimore:
Review requested -
@andlar 
@brianlindsy 
@kekey1

### DIFF
--- a/src/test/java/stepDefinitions/SEDTweaksSteps.java
+++ b/src/test/java/stepDefinitions/SEDTweaksSteps.java
@@ -1,8 +1,22 @@
 package stepDefinitions;
 
-import java.util.List;
-
+import org.openqa.selenium.By;
 import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.interactions.Actions;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.WebDriverWait;
+
+import cucumber.api.PendingException;
+import cucumber.api.java.en.Given;
+import cucumber.api.java.en.Then;
+import cucumber.api.java.en.When;
+import static org.junit.Assert.assertTrue;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+
 public class SEDTweaksSteps {
 	public WebDriver driver;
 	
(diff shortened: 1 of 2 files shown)


src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
 		
 	}
 	
+	@Given("^I am on a listing's details pages$")
+	public void i_am_on_a_listing_s_details_pages() throws Throwable {
```
andlar: Could this be parameterized? Something like "Given I am on the details page for <id> on <ENV>" where "id" & "ENV" could be used to pick and choose between different products (with ID) and different environments (with ENV)? That way we'd be able to write this function once, but use it many times?

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+	    driver.get("https://chpl.ahrqstg.org/#/product/8481");
+	    WebDriverWait wait = new WebDriverWait(driver, 20);
+		driver.manage().window().maximize();
+		String actualTitle = driver.getTitle();
+		System.out.println("page title is : "+actualTitle);
+	    
+	}
+
+	@When("^I look at the accordions$")
+	public void i_look_at_the_accordions() throws Throwable {
```
andlar: This doesn't actually do anything, right? It just "system.out"s a bunch of times? First, I think the "System.out"s should be removed from "complete" code; fine for testing, but shouldn't be there in a PR. Second, I think this step itself has no real purpose. I'd rather have the Scenario written without the step at all. I think every step should either _do_ or _assert_ something.

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+	    System.out.println("Accordion title:" +a4);
+	    String a5 = driver.findElement(By.cssSelector("div.panel:nth-child(6) > div:nth-child(1) > a:nth-child(1)")).getText();
+	    System.out.println("Accordion title:" +a5);
+	    String a6 = driver.findElement(By.cssSelector("div.panel:nth-child(7) > div:nth-child(1) > a:nth-child(1)")).getText();
+	    System.out.println("Accordion title:" +a6);
+	    
+	    
+	}
+
+	@Then("^I should see them in the following order: Criteria, CQMs, SED, G(\\d+)/G(\\d+), Surveillance, Additional Info$")
```
andlar: The "g1/g2" stuff isn't really a regex. It's literally "G1/G2", and something like "G9/G3" would be wrong

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+		assertTrue(driver.findElement(By.cssSelector("div.panel:nth-child(2) > div:nth-child(1) > a:nth-child(1)")).getText().contains("Certification Criteria"));
+		assertTrue(driver.findElement(By.cssSelector("div.panel:nth-child(3) > div:nth-child(1) > a:nth-child(1)")).getText().contains("Clinical Quality Measures"));
+		assertTrue(driver.findElement(By.cssSelector("div.panel:nth-child(4) > div:nth-child(1) > a:nth-child(1)")).getText().contains("Safety Enhanced Design (SED)"));
+		assertTrue(driver.findElement(By.cssSelector("div.panel:nth-child(5) > div:nth-child(1) > a:nth-child(1)")).getText().contains("Successfully Tested G1/G2 Measures"));
+		assertTrue(driver.findElement(By.cssSelector("div.panel:nth-child(6) > div:nth-child(1) > a:nth-child(1)")).getText().contains("Surveillance Activities"));
+		assertTrue(driver.findElement(By.cssSelector("div.panel:nth-child(7) > div:nth-child(1) > a:nth-child(1)")).getText().contains("Additional Information"));
+		
+	}
+
+	@Given("^I am looking at a listing's SED details$")
```
andlar: This doesn't actually do what it says. It goes to a page, but not to the SED Details panel on that Listing

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+		WebDriverWait wait = new WebDriverWait(driver, 20);
+		wait.until(ExpectedConditions.presenceOfElementLocated(By.cssSelector("div.panel:nth-child(4) > div:nth-child(1) > a:nth-child(1)")));
+		
+	}
+
+	@When("^I look at the SED Intended User Description$")
+	public void i_look_at_the_SED_INtended_User_Description() throws Throwable {
+		
+	
+		Thread.sleep(5000);
```
andlar: Any time I see `Thread.sleep(5000);` or something like it I cringe. Is there no other way to determine if a page is loaded? Arbitrary length sleeps could go wrong in so many ways. That's one thing I liked about the JS selenium / protractor combination; it "understands" Angular, so can _know_ when the page is loaded

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+
+	@Given("^I am looking at a listing's SED details$")
+	public void i_am_looking_at_a_listing_s_SED_details() throws Throwable {
+		driver.get("https://chpl.ahrqstg.org/#/product/8481");
+		WebDriverWait wait = new WebDriverWait(driver, 20);
+		wait.until(ExpectedConditions.presenceOfElementLocated(By.cssSelector("div.panel:nth-child(4) > div:nth-child(1) > a:nth-child(1)")));
+		
+	}
+
+	@When("^I look at the SED Intended User Description$")
```
andlar: What does this step do? It looks like it clicks a couple of things, but then system.outs something or line 94, which does what? If it's supposed to be opening the SED panel, then the step should be something like "When I open the SED Panel of the accordion"

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+		    System.out.println("No Certification Criteria were tested for SED.");
+		else
+			driver.getPageSource().contains("Description of Intended Users");
+	}
+
+	@Then("^the header should be named Description of Intended Users$")
+	public void the_header_should_be_named_Description_of_Intended_Users() throws Throwable {
+	    
+		assertTrue(
+				driver.getPageSource().contains("Description of Intended Users"));
```
andlar: Will this be true if the header is anywhere on the page? Should it be adjusted to find a particular thing, whether by class, id, or page structure?

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+			
+	}
+
+	@When("^I look at the 'SED Testing Completion Date'$")
```
andlar: What does this do? I don't see either an action or an assertion...

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+
+	@Then("^the header should be named 'Date SED Testing was Completed'$")
+	public void the_header_should_be_named_Date_SED_Testing_was_Completed() throws Throwable {
+	    assertTrue(
+				driver.getPageSource().contains("Date SED Testing was Completed"));
+	}
+
+	@When("^I look at the download button$")
+	public void i_look_at_the_download_button() throws Throwable {
+	   driver.findElement(By.cssSelector(".panel-body > div:nth-child(1) > span:nth-child(1) > span:nth-child(5) > table:nth-child(2) > tfoot:nth-child(2) > tr:nth-child(1) > td:nth-child(1) > button:nth-child(1)")).getTagName();
```
andlar: This looks like an opportune time to do a findById and make the UI developer actually put an ID on that button

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+	}
+
+	@Then("^the button should be named 'Download SED Details'$")
+	public void the_button_should_be_named_Download_SED_Details() throws Throwable {
+		assertTrue(
+				driver.getPageSource().contains("Download SED Details"));
+	}
+
+	@When("^I look at the associated criteria$")
+	public void i_look_at_the_associated_criteria() throws Throwable {
```
andlar: This step isn't looking at criteria. It's loading a page. It should be either written to look at criteria or subsumed under the "go to a Listing Details page" step from earlier

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+	@Then("^the criteria number should also include the criteria title$")
+	public void the_criteria_number_should_also_include_the_criteria_title() throws Throwable {
+		Thread.sleep(5000);
+		WebElement expanded = driver.findElement(By.cssSelector("div.panel:nth-child(2) > div:nth-child(1) > a:nth-child(1)"));
+		expanded.click();
+					
+		WebElement expanded1 = driver.findElement(By.cssSelector("div.panel:nth-child(4) > div:nth-child(1) > a:nth-child(1)"));
+		expanded1.click();
+		
+		 WebElement Table_1 = driver.findElement(By.xpath("//*[@id='mainContent']/div[2]/div[2]/div/div/div[3]/div[2]/div/span/span[4]/table/thead/tr/th[1]"));
```
andlar, line 135: Can/should we start putting IDs on tables? Or something? This xpath is way too deep, and that means it's very fragile. A single added (or removed) span anywhere in here and the test breaks.

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+		  for (WebElement col : Cols) {
+              System.out.print(col.getText() + "\t");
+		  }
+		  System.out.println();
+		 }
+		}
+		
+	
+
+	@Given("^I am looking at listing's SED details$")
```
andlar: How is this different from the function at line 73?

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -12,7 +26,171 @@ public SEDTweaksSteps()
...
+	public void i_am_looking_at_listing_s_SED_details() throws Throwable {
+	    driver.get("https://chpl.ahrqstg.org/#/product/8482");
+	    WebDriverWait wait = new WebDriverWait(driver, 20);
+		wait.until(ExpectedConditions.presenceOfElementLocated(By.cssSelector("div.panel:nth-child(4) > div:nth-child(1) > a:nth-child(1)")));
+		
+	}
+
+	@Given("^present listing doesn't have SED$")
```
andlar: Maybe we should use one Step for "navigate to a listing" and have that be parameterizable, and have all of the steps where we need to get to a particular type of Listing leverage that step

src/test/resources/Features/SEDTweaks.feature
```diff
@@ -1,14 +1,14 @@
 Feature: SED Display in product details
 This test is to verify text changes and other display updates to SED section 
 
-Scenario:
+Scenario: Verify SED details
 Given I am on a listing's details pages
 When I look at the accordions
 Then I should see them in the following order: Criteria, CQMs, SED, G1/G2, Surveillance, Additional Info
 
 Given I am looking at a listing's SED details
```
andlar: Each of these "Given When Then"s  should have a Scenario title

andlar (changes requested):
There are a lot of comments here. I don't mean to be picky, and maybe it's not worth addressing all of these now, but I think we should try to get started on our AQA stuff down a path that pushes us in the right direction. Am up for a discussion about that...

kekey1 (dismissed):

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: I think this is a nice-to-have. A future AQA task might be to add environment properties to parameterize all calls.

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: Agree with removing the System.outs and making sure each test is checking something with an assert function or similar.

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: Agree with this for a fix.

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: Agree with making this open the accordion and check for some value

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: Can you listen for a "loaded" or "initialized" event on some field? Maybe this could be figured out at a later time, but I agree that waiting 5 seconds is not a reliable way to test something for in the longer term.

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: If this is succeeding for now, I'm not sure if it needs to be fixed for the PR

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: Agree with making sure all tests are asserting something

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: Yeah! Make the UI developer do stuff! I would say not necessary for the PR if it's working right now but good practice going forward if you find you have complicated selectors to get to an element.

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: Agree with moving the Then check to the 'go to a Listing Details page' from earlier.

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: Probably not necessary to change for the PR. As AQA goes along, Ashwini should make a note to ask for IDs anytime she has to write expressions like that

src/test/java/stepDefinitions/SEDTweaksSteps.java
kekey1: Agree with the idea behind this comment. Let's consolidate the "navigate to a listing" functionality into one place that can have a parameter passed in but let's do it in the next iteration of AQA development. Next time Ashwini needs to load a listing page, figure out how to do that with parameters and circle back around to this code.

kekey1 (changes requested):
Tried to comment on each of Andrew's comments if I thought something needed to be fixed for the PR.

andlar (approved):

brianlindsy (approved):

kekey1 (approved):

PR #5, opened 2017-12-26
OCD-1940-script to verify text after fuzzy match update

ashwinimore:
Please review PR:
@andlar 
@brianlindsy 
@kekey1


kekey1 (approved):

brianlindsy (approved):

src/test/java/pageObjects/chpldownload_page.java
```diff
@@ -0,0 +1,29 @@
...
+import org.openqa.selenium.WebElement;
+
+import stepDefinitions.hooks;
+
+public class chpldownload_page {
+	
+	private WebDriver driver;
+	private static WebElement element = null;
+	
+	public static void print_page_title(WebDriver driver) {
```
andlar: This looks like the kind of thing that should be in some "helper" class. It's not specific to the Download page, but could be used on any page and just return the title as a String for tests to use to verify stuff

src/test/java/pageObjects/chpldownload_page.java
```diff
@@ -0,0 +1,29 @@
...
+		driver = hooks.driver;
+		driver.manage().window().maximize();
+		String actualTitle = driver.getTitle();
+		System.out.println("page title is : "+actualTitle);
```
andlar: We shouldn't have `System.out.println` in _any_ of our real code (maybe; there are never any absolutes, including this one, but still...). It's great for debugging, but a Page Object Class should just return things

src/test/java/pageObjects/listingDetails_page.java
```diff
@@ -1,58 +1,150 @@
...
+import org.openqa.selenium.support.ui.WebDriverWait;
+
+import stepDefinitions.hooks;
+
+public class listingDetails_page {
+	
+	private WebDriver driver;
+	private static WebElement element = null;
+	
+	public static void load_listings_details_page(WebDriver driver) {
```
andlar: The indentation on this Class is weird. I think we should get @kekey1 's help with setting up that checkstyle stuff and/or auto fixing as much of the rules we have in CHPL as possible

src/test/java/pageObjects/listingDetails_page.java
```diff
@@ -1,58 +1,150 @@
...
+			driver = hooks.driver;
+			driver.manage().window().maximize();
+			String actualTitle = driver.getTitle();
+			System.out.println("page title is : "+actualTitle);
+	}
+		public static void sed_details_accordion_open(WebDriver driver){
+		WebDriverWait wait = new WebDriverWait(driver, 20);
+		wait.until(ExpectedConditions.presenceOfElementLocated(By.id("details-sed")));
+		
+		// to close certification criteria accordion that's open by default
```
andlar, line 21: This isn't necessary, as opening any of the accordions automatically closes all of the others

src/test/java/pageObjects/listingDetails_page.java
```diff
@@ -1,58 +1,150 @@
...
+		listingDetails_page.certification_criteria_accordion(driver);
+		
+		//to open SED accordion			
+		listingDetails_page.SED_accordion(driver);
+	}
+	
+	public static WebElement edit_certified_product_link(WebDriver driver){
+		element = driver.findElement(By.linkText("Return to search results"));
```
andlar: "Return to search results" doesn't match `edit_certified_product_link`

src/test/java/pageObjects/listingDetails_page.java
```diff
@@ -1,58 +1,150 @@
...
+		element.click();
+		return element;
+	}
+	
+	public static WebElement SED_criteria_table(WebDriver driver){
+		element = driver.findElement(By.xpath("//*[@id='mainContent']/div[2]/div[2]/div/div/div[3]/div[2]/div/span/span[4]/table/thead/tr/th[1]"));
```
andlar: I think I added an id to that table; can you check that? It would make this a simpler xpath

src/test/java/pageObjects/listingDetails_page.java
```diff
@@ -1,58 +1,150 @@
...
+		return element;
+	} 
+	
+	public static WebElement certification_criteria_a1_view_details(WebDriver driver){
+		element = driver.findElement(By.xpath("//*[@id=\"mainContent\"]/div[2]/div[2]/div/div/div[1]/div[2]/div[1]/div/div/span[2]/a"));
```
andlar: Similarly with these; isn't there something like a `certification_results_panel` ID somewhere we should start from?

src/test/java/pageObjects/listingDetails_page.java
```diff
@@ -1,58 +1,150 @@
...
+	public static WebElement certification_criteria_g4_view_details(WebDriver driver){
+		element = driver.findElement(By.xpath("//*[@id=\"panel-certification-criteria\"]/div[53]/div/div/span[2]/a"));
+		Actions action = new Actions(driver);
+		action.moveToElement(element).click().perform();
+		return element;
+	}
+	
+	public static WebElement QMS_standard_text(WebDriver driver){
+		String actualString = driver.findElement(By.xpath("//*[@id=\"panel-certification-criteria\"]/div[53]/div/div/div/table/tbody/tr[3]/td[2]/ul/li")).getText();
+		assertTrue(actualString.contains("Standard: ISO9001:2015"));
```
andlar: Page Objects shouldn't have any assertions

src/test/java/pageObjects/listingDetails_page.java
```diff
@@ -1,58 +1,150 @@
...
+	public static WebElement certification_criteria_g5_view_details(WebDriver driver){
+		element = driver.findElement(By.xpath("//*[@id=\"panel-certification-criteria\"]/div[54]/div/div/span[2]/a"));
+		Actions action = new Actions(driver);
+		action.moveToElement(element).click().perform();
+		return element;
+	}
+	
+	public static WebElement Acceessibility_standard_text(WebDriver driver){
+		String actualString = driver.findElement(By.xpath("//*[@id=\"panel-certification-criteria\"]/div[54]/div/div/div/table/tbody/tr[3]/td[2]/ul/li[1]")).getText();
+		assertTrue(actualString.contains("Other - WCAG 2.0"));
```
andlar: Page Objects shouldn't have any assertions

src/test/java/pageObjects/listingDetails_page.java
```diff
@@ -1,58 +1,150 @@
...
+		return element;
+	}
+	public static WebElement Acceessibility_standard_text2(WebDriver driver){
+		String actualString = driver.findElement(By.xpath("//*[@id=\"panel-certification-criteria\"]/div[54]/div/div/div/table/tbody/tr[3]/td[2]/ul/li[2]")).getText();
+		assertTrue(actualString.contains("Section 508 of the Rehabilitation Act"));
```
andlar: Page Objects shouldn't have any assertions

src/test/java/pageObjects/listingDetails_page.java
```diff
@@ -1,58 +1,150 @@
...
+	
+	public static WebElement SED_accordion(WebDriver driver){
+		element = driver.findElement(By.id("details-sed"));
+		element.click();
+		return element;
+	}
+	
+	public static WebElement UCD_process_text(WebDriver driver){
+		String actualString = driver.findElement(By.xpath("//*[@id=\"sed-ucd-processes-table\"]/tbody/tr/td[2]")).getText();
+		assertTrue(actualString.contains("NISTIR 7741"));
```
andlar: Page Objects shouldn't have any assertions

src/test/java/pageObjects/search_page.java
```diff
@@ -0,0 +1,39 @@
...
+	private WebDriver driver;
+	private static WebElement element = null;
+	
+	public static WebElement search_field(WebDriver driver){
+		element = driver.findElement(By.id("searchField"));
+		return element;
+	}
+	public static WebElement search_results_chpl_id(WebDriver driver){
+		String actualString = driver.findElement(By.xpath("//*[@id='resultsSection']/div[2]/div/table/tbody/tr/td[6]")).getText();
+		assertTrue(actualString.contains("14.07.07.1221.FIA1.16.01.1.170608"));
```
andlar: Page Objects shouldn't have any assertions

src/test/java/pageObjects/search_page.java
```diff
@@ -0,0 +1,39 @@
...
+	
+	public static WebElement details_link(WebDriver driver){
+		element = driver.findElement(By.xpath("//*[@id=\'resultsSection\']/div[2]/div/table/tbody/tr/td[8]/div/div[1]/a"));
+		element.click();
+		return element;
+	}
+	
+	public static WebElement disclosure_url(WebDriver driver){
+		String actualString = driver.findElement(By.xpath("//*[@id=\"mainContent\"]/div[2]/div[1]/span/div[2]/a")).getText();
+		assertTrue(actualString.contains("cehrt-disclosure-information"));
```
andlar: Page Objects shouldn't have any assertions

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -1,107 +1,107 @@
...
+ 	
+	@Given("^I am on the Details page of Listing \"([^\"]*)\"$")
+	public void i_am_on_a_listing_s_details_pages(String arg1) throws Throwable {
+		driver.get("https://chpl.ahrqstg.org/#/product/" + arg1);
+		WebDriverWait wait = new WebDriverWait(driver, 20);
+		listingDetails_page.load_listings_details_page(driver); 	    
+	}
+
+	
+	@Then("^I should see the accordions in the following order: Criteria, CQMs, SED, G1/G2, Surveillance, Additional Info$")
```
andlar: This test doesn't test what it says it tests. It says it's testing ordering, but it's actually testing existence of. I think the reason for the original test was to test ordering, so the logic below should probably change. It wouldn't be wrong to _also_ have a test that tests the existence of, especially as different listings will have different sets of "existing" panels

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -1,107 +1,107 @@
...
+
+	@When("^I open SED details accordion$")
+	public void i_open_SED_accordion() throws Throwable {
+		listingDetails_page.sed_details_accordion_open(driver);
+	}
+	
+	@Then("^usability report text should read as 'Full Usability Report'$")
+	public void usability_report_title_reads_as_Full_Usability_Report() throws Throwable {
+		
+	assertTrue(driver.getPageSource().contains("Full Usability Report"));
```
andlar: We want to use Page Objects to get to the items, right? So this `assert` should be something like `listingDetails_page.getSedUsabilityReportTitle().contains("Full Usability Report"));` and the Page Object should add that method (`getSedUsabilityReportTitle`) to match it

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -1,107 +1,107 @@
...
+	}
+
+	@Then("^intended user description header should show as Description of Intended Users$")
+	public void the_header_should_be_named_Description_of_Intended_Users() throws Throwable {
+	    
+	assertTrue(driver.getPageSource().contains("Description of Intended Users"));
```
andlar: Same as above

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -1,107 +1,107 @@
...
+			
+	}
+
+	@Then("^the header name should read as 'Date SED Testing was Completed'$")
+	public void the_header_should_be_named_as_Date_SED_Testing_was_Completed() throws Throwable {
+		
+	assertTrue(driver.getPageSource().contains("Date SED Testing was Completed"));
```
andlar: Same as above

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -1,107 +1,107 @@
...
+	}
+
+	@Then("^the download button title should read as 'Download SED Details'$")
+	public void the_download_button_title_should_read_as_Download_SED_Details() throws Throwable {
+		
+	assertTrue(driver.getPageSource().contains("Download SED Details"));
```
andlar: Same as above

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -1,107 +1,107 @@
...
+	}
+	
+	@Then("^the criteria number should also include the criteria title$")
+	public void the_criteria_number_should_also_include_the_criteria_title() throws Throwable {
+		assertTrue(driver.getPageSource().contains("170.315 (a)(1): Computerized Provider Order Entry (CPOE) - Medications"));
```
andlar: Similar to above. This should call some sort of method in the PageObject that returns an Array of criteria titles, and then assert that various criteria w/titles are in that array. And probably that some other ones aren't; both criteria the Listing doesn't attest to, and criteria _without_ the title. E.g., `assertTrue(criteriaTitles.not.contains("170.315 (a)(3)"))`; where `criteriaTitles` is defined above as something like `String [] criteriaTitles = listingDetails.getCriteriaTitles()"`

src/test/java/stepDefinitions/chpldownloadSteps.java
```diff
@@ -1,146 +1,145 @@
...
+	
+	@Then("^Download the CHPL page loads successfully and page title shows correct$")
+	public void download_the_CHPL_page_loads_successfully_and_page_title_shows_correct() throws Throwable {
+	    
+		chpldownload_page.print_page_title(driver);
+	}
+
+	@Given("^user is not logged in$")
+	public void user_is_not_logged_in() throws Throwable {
+	System.out.println("user is not logged in");
```
andlar: This `@Given` will need to be updated once log in works; it should check to see if the user is logged in, and if it is, should figure out how to click the "Log Out" button. Probably doesn't need to be done now, but eventually

src/test/java/stepDefinitions/chpldownloadSteps.java
```diff
@@ -1,146 +1,145 @@
...
+	driver.manage().timeouts().implicitlyWait(60, TimeUnit.SECONDS);
+	}
+	
+	@When("^user clicks on download file box$")
+	public void user_clicks_on_download_file_box() throws Throwable {
```
andlar: This Step doesn't do anything. The PageObject automatically clicks on the select box, so when, in the "I should see n Items" Step below, the select box is found, it's been clicked already. This Step and the next should be refactored, and the entire test can probably we changed to something like: `Given I am on the Download Page And I am not logged in Then I should see n options`

src/test/java/stepDefinitions/chpldownloadSteps.java
```diff
@@ -1,146 +1,145 @@
...
+}
+		
+	
+	@Then("^user sees 7 download files")
+	public void user_sees_7_download_files() throws Throwable {
+		
+		WebElement element=chpldownload_page.downloadselect_list(driver);
+		Select select=new Select(element);
+		List<WebElement> list=select.getOptions();
+		System.out.println("Number of download files: " +list.size());
```
andlar: Shouldn't this be an `assertTrue`? It's a `@Then` Step, which means it should be doing assertions, right?

src/test/java/stepDefinitions/chpldownloadSteps.java
```diff
@@ -1,146 +1,145 @@
...
+		dropdown.selectByVisibleText("2015 edition products (xml)");
+	}
+	
+	@Then("^definition file shows based on download file selection$")
+	public void definitio_file_shows_based_on_selection () throws Throwable{
+	
+		WebElement selectBox2 = chpldownload_page.downloadselect_list(driver);
+		Select definition = new Select(selectBox2);
+		List<WebElement> definitionOptions = definition.getOptions();
+		System.out.println("Number of definition files: " +definitionOptions.size());
```
andlar: Also needs an assertion

src/test/java/stepDefinitions/chpldownloadSteps.java
```diff
@@ -1,146 +1,145 @@
...
+		Thread.sleep(100);
+		// This part is currently not logging in user; login button click() is not working
+		driver.findElement(By.id("username")).sendKeys("username");
+		driver.findElement(By.id("password")).sendKeys("password");
+		driver.findElement(By.xpath("//*[@id='admin']/li/div/form/button[1]")).click();
+		  
+	}
+
+	@When("^User checks download file box$")
+	public void user_checks_download_file_box() throws Throwable {
```
andlar: Does this do anything?

src/test/java/stepDefinitions/chpldownloadSteps.java
```diff
@@ -1,146 +1,145 @@
...
+}
+	
+	@Then("^user sees 8 download files")
+	public void user_sees_8_download_files() throws Throwable {
+		
+		WebElement element=chpldownload_page.downloadselect_list(driver);
+		Select select=new Select(element);
+		List<WebElement> list=select.getOptions();
+		System.out.println("Number of download files: " +list.size());
+		for(int i=0;i<list.size();i++)
```
andlar: Needs assertion

src/test/java/stepDefinitions/disclosureurlSteps.java
```diff
@@ -0,0 +1,45 @@
...
+public class disclosureurlSteps {
+	public WebDriver driver;
+	private WebElement element;
+	public disclosureurlSteps(){
+		driver = hooks.driver;
+	}
+	
+	@Given("^I am on listing details page of listing with CHPL ID \"(.*)\"$")
+	public void i_am_on_listing_details_page_of_listing_with_CHPL_ID(String CHPL_ID) throws Throwable {
+		driver.get("https://chpl.ahrqstg.org/#/search");
```
andlar, line 24: Why doesn't this just go straight to the details page? This step depends on the Search page, which means it's fragile

src/test/java/stepDefinitions/disclosureurlSteps.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+	
+	@When("^I look at Mandatory Disclosure URL$")
+	public void i_look_at_Mandatory_Disclosure_URL() throws Throwable {
+		driver.findElement(By.xpath("//*[@id=\"mainContent\"]/div[2]/div[1]/span/div[2]/h2"));		   
+	}
+
+	@Then("^it should show updated url$")
+	public void it_should_show_updated_url() throws Throwable {
+		search_page.disclosure_url(driver);
```
andlar: There should be an assertion here

src/test/java/stepDefinitions/disclosureurlSteps.java
```diff
@@ -0,0 +1,45 @@
...
+		driver.get("https://chpl.ahrqstg.org/#/search");
+		search_page.search_field(driver).sendKeys(CHPL_ID);
+		WebDriverWait wait = new WebDriverWait(driver, 20);
+		search_page.details_link(driver);
+		search_page.disclosure_url(driver);		  
+	}
+	
+	@When("^I look at Mandatory Disclosure URL$")
+	public void i_look_at_Mandatory_Disclosure_URL() throws Throwable {
+		driver.findElement(By.xpath("//*[@id=\"mainContent\"]/div[2]/div[1]/span/div[2]/h2"));
```
andlar: I think the "I look at..." paradigm isn't right; the "I look at `<a thing>`" that's written in Jira should probably be interpreted as "Build a part of the Page Object that allows the Steps to reference the `<a thing>`"

src/test/java/stepDefinitions/fuzzymatchtextSteps.java
```diff
@@ -0,0 +1,63 @@
...
+@Given("^I am on listing details page \"([^\"]*)\"$")
+public void i_am_on_listing_details_page(String arg1) throws Throwable {
+	driver.get("https://chpl.ahrqdev.org/#/product/" + arg1);
+	WebDriverWait wait = new WebDriverWait(driver, 20);
+	listingDetails_page.load_listings_details_page(driver); 	    
+}
+
+@When("^I look at criteria details for criteria g4$")
+public void i_look_at_criteria_details_g4() throws Throwable {
+	listingDetails_page.certification_criteria_g4_view_details(driver);
```
andlar: Similar to above wrt the "I look at `<a thing>`" paradigm

src/test/java/stepDefinitions/fuzzymatchtextSteps.java
```diff
@@ -0,0 +1,63 @@
...
+}
+
+@Then("^QMS Standard should display updated 'Standard: ISO9001:2015' text$")
+public void QMS_standard_displays_correct_text() throws Throwable {
+	listingDetails_page.QMS_standard_text(driver);
```
andlar: Similar to above `assertTrue`s

src/test/java/stepDefinitions/SEDTweaksSteps.java
```diff
@@ -1,107 +1,107 @@
...
+		assertTrue(driver.getPageSource().contains("170.315 (a)(6): Problem List"));
+		assertTrue(driver.getPageSource().contains("170.315 (a)(8): Medication Allergy List"));
+		assertTrue(driver.getPageSource().contains("170.315 (a)(14): Implantable Device List"));
+		assertTrue(driver.getPageSource().contains("170.315 (b)(3): Electronic Prescribing"));
+		assertTrue(driver.getPageSource().contains("170.315 (a)(7): Medication List"));
+	}
+		
+	
+	@Given("^I am on the details page of listing \"([^\"]*)\"$")
+	public void i_open_SED_accordion(String arg3) throws Throwable {
```
andlar: Why `arg3`? For that matter, why `arg1` and `arg2`? They should be meaningful names; for most of them I'd guess they should be something like `chplId`, though my guess (without looking) is the `arg1/arg2` will be `username/password`

andlar (changes requested):

andlar (approved):

PR #10, opened 2018-02-12
CMS ID, fuzzy matching tests and other chnages to reorder and fix some warnings

ashwinimore:
Review requested
@andlar 
@brianlindsy


src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
```diff
@@ -0,0 +1,98 @@
...
+import gov.healthit.chpl.aqa.pageObjects.CMSidReverseLookupPage;
+import gov.healthit.chpl.aqa.pageObjects.SearchPage;
+
+/**
+ * Class CMSIDsLookupSteps definition.
+ */
+public class CMSIDsLookupSteps {
+
+    private WebDriver driver;
+    private static final int DELAY = 40;
```
andlar: This is a "TIMEOUT", not a "DELAY", right?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
```diff
@@ -0,0 +1,98 @@
...
+    driver = Hooks.getDriver();
+    }
+    
+    /**
+     * To load CMS ID Reverse Lookup page 
+     */
+    
+    @Given("^I am an end-user on CMS ID Reverse Lookup page$")
+    public void i_am_an_end_user_on_CMS_ID_Reverse_Lookup_page() {
+        driver.get("https://chpl.ahrqstg.org/#/resources/cms_lookup");
```
andlar: Please look at the SED Steps and implement the part with the url that's brought in from the command line

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
```diff
@@ -0,0 +1,98 @@
...
+        assertTrue(CMSidReverseLookupPage.CertIdResultsTable_CHPLIdcolumn(driver).getText().contains(chplId));
+    }
+    
+    /**
+     * Loads CHPL home/search page  
+     */
+    
+    @Given("^I am an end user on CHPL home page$")
+    public void i_am_an_end_user() throws Throwable {
+        driver.get("https://chpl.ahrqstg.org/#/search");
```
andlar: Please use that "url" variable as from the SED Steps

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
```diff
@@ -0,0 +1,98 @@
...
+        WebDriverWait wait = new WebDriverWait(driver, DELAY);
+        driver.manage().window().maximize();
```
andlar: I think there's something about the maximize call that breaks stuff. If it's not necessary for the step please remove it

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/FuzzyMatchTextSteps.java
```diff
@@ -25,7 +25,7 @@ public FuzzyMatchTextSteps() {
 
     @Given("^I am on listing details page \"([^\"]*)\"$")
```
andlar, line 32: I think this step is a duplicate of one of the ones in a different step file, but with a slightly different name. Shouldn't re-invent that effort

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -17,6 +17,8 @@
 
     @Before
     public void openBrowser() {
+        String exePath = "C:\\Users\\amore\\eclipse-workspace\\chpl-testautomation\\chromedriver_win32\\chromedriver.exe";
```
andlar: This will break any other user's environment, as well as Jenkins. You should have your chromedriver directory in your $PATH or %PATH% environment variable and remove this stuff

andlar (changes requested):
For the Feature files, can you name then with the as "OCD-nnnn.feature" where the nnnn is the parent task of whatever Feature you're testing?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
```diff
@@ -56,43 +64,39 @@ public void i_should_see_the_listings_that_make_up_the_CMS_ID(final String chplI
     
     @Given("^I am an end user on CHPL home page$")
     public void i_am_an_end_user() throws Throwable {
-        driver.get("https://chpl.ahrqstg.org/#/search");
-        WebDriverWait wait = new WebDriverWait(driver, DELAY);
-        driver.manage().window().maximize();
+        driver.get(url + "#/search");
+        WebDriverWait wait = new WebDriverWait(driver, TIMEOUT);
```
andlar: This wait doesn't do anything, right? I mean, it won't actually wait until you do some sort of "wait.until()" call

src/test/resources/Features/CMSIDsLookup.feature
```diff
@@ -0,0 +1,19 @@
...
+      |0015H8GK6K0ZZB2|15.04.04.2891.Sunr.07.01.1.171201|
+      			
+Scenario Outline: If a user loads their cms widget with any of these sets of listings, and generates a CMS ID then it should return the newer CMS ID
+	Given I am an end user on CHPL home page
+	When I Generate CMS ID using CMS ID widget for same listings as in result of reverse lookup
+	Then the generated CMS ID should be "<CMS_ID>"
+	Examples:
+      |CMS_ID|	
+      |0015H8GK6K0ZZB2|
+
```
andlar: This should be something where you pass in a comma separated list of CHPL IDs and the CMS ID that should be generated from those IDs, something like:
`|CHPL-0000,CHPL-3232|14ds0sfdfd|`
and then the step that does the "add to the cms widget" should loop through the CHPL IDs

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/FuzzyMatchTextSteps.java
```diff
@@ -25,7 +25,7 @@ public FuzzyMatchTextSteps() {
...
     public void iAmOnListingDetailsPage(final String arg1) throws Throwable {
-        driver.get("https://chpl.ahrqstg.org/#/product/" + arg1);
+        driver.get("https://chpl.ahrqdev.org/#/product/" + arg1);
```
andlar: Please update to use the URL stuff, or get rid of this step entirely and use the other one that does the same thing

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Loads a listing. First searches for listing, then adds Cert Id to CMS creator.
+     * Waits after adding each of the Cert Ids, then generates EHR certification Id.
+     */
+    
+    @When("^I Generate CMS ID using CMS ID widget for same listings as in result of reverse lookup$")
+    public void i_Generate_CMS_ID_using_CMS_ID_widget_for_same_listings_as_in_result_of_reverse_lookup() {
+        SearchPage.searchField(driver).sendKeys("14.07.07.2452.VEI1.01.01.0.161111");
```
andlar: This Step should take a chunk of CHPL IDs and loop through them to add them to the CMS ID Widget

andlar (changes requested):
Please also change the filenames for the Feature files to reference the OCD number of the main ticket

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
ashwinimore: Each of the Cert Id element has unique 'id' (associated database id) that requires click event (unlike generic link used in disclosureurl test). I tried using datatable, lists and examples without any luck. The test allows adding bunch of CHPL IDs but isn't as good of an option as it should be. I'll work on improving that.

src/test/resources/Features/CMSIDsLookup.feature
ashwinimore: Comma separated list sends as-is value in sendkeys (CHPL-0000,CHPL-3232 as input in search box in case of above example), I took a different approach.

src/test/java/gov/healthit/chpl/aqa/pageObjects/ListingDetailsPage.java
```diff
@@ -14,35 +14,23 @@
...
+        element = driver.findElement(By.id("details-additional-information"));
         return element;
     }
-
-    public static WebElement sEDCriteriaTable(final WebDriver driver) {
-        element = driver.findElement(By.xpath("//*[@id=\"sed-ucd-processes-table\"]"));
+    
+    public static WebElement certificationCriteriaAccordion(final WebDriver driver) {
+        element = driver.findElement(By.id("details-certification-criteria"));
+        element.click();
```
andlar: Why does this execute the click? It would expand / contract the accordion element, right? If the method is only to return teh element, then I don't think it should do things with the element.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -21,21 +23,28 @@
 
     private WebDriver driver;
     private static final int DELAY = 20;
+    private String url = System.getProperty("url");
```
andlar: DELAY should be TIMEOUT

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -21,21 +23,28 @@
...
+            url = "http://localhost:3000/";
+        }
     }
 
     @Given("^user is on CHPL download page$")
     public void userIsOnCHPLHome() throws Throwable {
-        driver.get("https://chpl.ahrqdev.org/#/resources/download");
+        driver.get(url + "#/resources/download");
         WebDriverWait wait = new WebDriverWait(driver, DELAY);
         assertTrue(driver.getTitle().contains("Download the CHPL"));
```
andlar: The "wait" isn't being used, and assertTrue shouldn't be in an @Given step

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -21,21 +23,28 @@
...
     }
 
     @When("^user is not logged in$")
     public void userIsNotLoggedIn() throws Throwable {
-        driver.get("https://chpl.ahrqdev.org/#/resources/download");
+        driver.get(url + "#/resources/download");
```
andlar: User is not logged in shouldn't navigate anywhere. For that matter, I think it might be better to just take out this login stuff entirely until we can figure out a way to actually do it

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -70,7 +79,7 @@ public void definitionFileShowsBasedOnSelection() throws Throwable {
     @Given("^user is logged in with \"([^\"]*)\" and \"([^\"]*)\"$")
     public void userIsLoggedInWithAnd(final String arg1, final String arg2) throws Throwable {
 
-        driver.get("https://chpl.ahrqdev.org/#/resources/download");
+        driver.get(url + "#/resources/download");
```
andlar: Logging in shouldn't change the URL of where the test is

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/DisclosureUrlSteps.java
```diff
@@ -50,7 +50,7 @@ public void iAmOnListingDetailsPageOfListingWithCHPLID(final String chplId) {
      */
     @Then("^it should show updated url$")
     public void itShouldShowUpdatedUrl() {
-        String actualString = SearchPage.disclosureUrl(driver).getText();
+        String actualString = ListingDetailsPage.disclosureUrl(driver).getText();
```
andlar, line 53: Should pull in from upstream. This test has changed to have the actual text of the disclosure URL

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/FuzzyMatchTextSteps.java
```diff
@@ -18,14 +20,21 @@
...
+     */
     public FuzzyMatchTextSteps() {
         driver = Hooks.getDriver();
+        if (StringUtils.isEmpty(url)) {
+            url = "http://localhost:3000/";
+        }
     }
 
     @Given("^I am on listing details page \"([^\"]*)\"$")
     public void iAmOnListingDetailsPage(final String arg1) throws Throwable {
```
andlar, line 33: If you write the Step in the Feature file as `@Given("^I am on the Details page of Listing \"([^\"]*)\"$")` then this Step should use the Step from the SEDTweaksStep.java file, and we won't have virtually identical steps

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/LoginSteps.java
```diff
@@ -46,7 +50,7 @@ public void loginFormShowsInDropdown() throws Throwable {
 
     @Given("^User enters valid \"(.*)\" and \"(.*)\"$")
     public void userEntersValidUsernameAndPassword(final String username, final String password) throws Throwable {
-        driver.get("https://chpl.ahrqdev.org/#/resources/download");
+        driver.get(url + "#/resources/download");
```
andlar: Shouldn't change pages in a step that doesn't say anything about changing pages

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/OverviewTest.java
```diff
@@ -32,15 +42,14 @@ public void overviewPageLoadsSuccessfullyAndVerifyTitle() throws Throwable {
 
     @When("^User looks at ONC-ACB and ONC-ATL information table$")
     public void userLooksAtONCACBAndONCATLInformationTable() throws Throwable {
-        driver.get("https://chpl.ahrqstg.org/#/resources/overview");
-        System.out.println("ONC-ACB and ONC-ATL Information");
+        driver.get(url + "#/resources/overview");
```
andlar: I think this step is unnecessary. Should probably be removed from the Feature file. In either case, the "act of looking" isn't the same as navigating to another page

andlar (changes requested):
Not trying to be too picky, but we have some time with this project now when there's not a lot of other QA stuff to do, so I'd like to clean this up as much as we possibly can

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/LoginSteps.java
ashwinimore: Login feature needs to be re-written (it was one of the initial scripts before setting up page objects). I'd have to request new ids/objects and add a new page under page objects just for login. That way, I can reuse login feature in other step definitions, wherever applicable. I am going to remove current step definition file.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/DisclosureUrlSteps.java
ashwinimore: I think updated test is not yet on chpladmin/chpl-test, maybe the changes aren't merged yet? When I pull from upstream, everything shows up-to-date.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/FuzzyMatchTextSteps.java
ashwinimore: It is cucumber limitation by design, step definitions are glued to feature file and matched based on the regexp(that is why i wrote it that way) and having same step in features is resulting in 'duplicate' error. I am going to look for workaround and probably go for a separate stepdefinition file to define common steps, just not sure of it yet.

andlar (approved):
👍

PR #12, opened 2018-02-24
OCD-2023:Monitor that download files on CHPL download resources page …

ashwinimore:
Tests to verify download files show correct date on filenames to check file generation didn't fail at any point.
Please review.
@andlar 
@brianlindsy


andlar (approved):
This looks awesome. There are a couple of "diffs" that introduce whitespace issues, but nothing substantive.

PR #13, opened 2018-02-27
OCD-2081-Verify duplicate listings are deleted

ashwinimore:
Tests to verify 3 duplicate drummond listings are deleted.
Please review:
@andlar 
@brianlindsy 
@tmy1313


andlar (approved):
I'm marking this as approved because it looks technically correct. I do want to mention that today's overnight run, which had some new AQA tests in it, had an additional 200-ish checkstyle errors/warnings. I don't need to have those all resolved, and definitely don't need this PR to resolve all the previous errors/warnings, but it'd be nice to reduce the number of them on an on-going basis. Running "mvn site" at the root of the project will generate that report locally, and if you have time to do that it'd be appreciated

debug.log
```diff
@@ -0,0 +1,2 @@
+[0228/131714.612:ERROR:prune_crash_reports.cc(34)] PruneCrashReportDatabase: Failed to get pending reports
```
andlar: Probably didn't mean to commit this file

PR #16, opened 2018-03-07
updated DisclosureURL tests to add Regression and Acceptance test 

ashwinimore:
…added @Regression tag to features
Please review:
@andlar 
@tmy1313 
@brianlindsy


src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -1,27 +1,55 @@
...
-
 /**
  * Class ChplDownloadSteps definition.
  */
 public class ChplDownloadSteps {
 
     private WebDriver driver;
-    private static final int TIMEOUT = 30;
     private String url = System.getProperty("url");
+    private static String downloadPath = "\\filepath\\tempdnld";
```
andlar: For this to work on linux, we'll need to have the download path set up as a property value, the same way the url is

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -1,27 +1,55 @@
...
+    /**
+     * Move files from download directory to other directory.
+     * This is to ensure files are archived before cleaning the directory.
+     * @throws Exception - to handle IOException
+     */
+    public void movefiles() throws Exception {
+        String source = "\\filepath\\tempdnld";
+        File srcDir = new File(source);
+
+        String destination = "\\filepath\\tempdnld1";
```
andlar: This one should be a property value too

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -34,13 +62,32 @@ public ChplDownloadSteps() {
...
+        options.addArguments("disable-popup-blocking");
+
+        DesiredCapabilities cap = DesiredCapabilities.chrome();
+        cap.setCapability(ChromeOptions.CAPABILITY, chromeOptionsMap);
+        cap.setCapability(CapabilityType.ACCEPT_SSL_CERTS, true);
+        cap.setCapability(ChromeOptions.CAPABILITY, options);
+
+        driver = new ChromeDriver(cap);
         driver.get(url + "#/resources/download");
-        WebDriverWait wait = new WebDriverWait(driver, TIMEOUT);
```
andlar: Should bring this back. With the long load time of our pages I think it's a good idea to have the "load the page" step not return until the page is actually fully loaded

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -73,4 +119,322 @@ public void userSeesDownloadFiles(final String expectedLength) {
...
+     */
+
+    @When("^I download the 2015 edition products file$")
+    public void download2015editionProductsFile() throws Exception {
+        /**
+         * Clean all files from download directory so each time there is only latest
+         * download file to read.
+         */
+        this.movefiles();
+        FileUtils.cleanDirectory(dir);
```
andlar: This step has to happen before each of these "@When"s, right? Maybe it would make more sense for it to be in the @Given step?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -73,4 +119,322 @@ public void userSeesDownloadFiles(final String expectedLength) {
...
+     */
+    @Then("^the downloaded file shows chpl-2015-currentdate.xml filename$")
+    public void verify2015editionProductsFilename() {
+        /**
+         * Get latest download file in the directory. Keep checking until download is
+         * complete then read final filename.
+         */
+
+        String dwldFileName = "";
+        int loopBack = 0;
```
andlar: Using an `int` here is weird. It would make more sense (and be more readable) if it was a boolean. Something like `boolean fileFound = false;` then in the "if" block below you can just do `if finalName.equals: fileFound = true`, and you won't need the else at all.

Similarly with the other instances of this pattern below.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplSearchSteps.java
```diff
@@ -1,52 +1,54 @@
...
-   }
+           }
+    }
+    /**
+     * Get user to CHPL search page.
+     */
+    @Given("^I am on CHPL search page$")
+    public void iAamOnCHPLSearchPage() {
+        driver.get(url);
+    }
```
andlar: Should have something here that waits until something on the page exists, otherwise the next step might start before it can work

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplSearchSteps.java
```diff
@@ -1,52 +1,54 @@
...
+     * Assert message when no results found.
      */
-    @Then("^the Mandatory Disclosures URL field should show Disclosures URL \"(.*)\"$")
-    public void verifyMandatoryDisclosureurlValue(final String urlValue) {
-        String actualText = ListingDetailsPage.disclosureUrl(driver).getText();
-        assertTrue(actualText.contains(urlValue), "Expect " + urlValue + " to be found in " + actualText);
-   }
+    @Then("^the search page shows 'No results found' message$")
+    public void verifyMessage() {
+        assertTrue(SearchPage.noResultsFound(driver).getText().contains("No results found"));
```
andlar: With the "assertTrue" it's really helpful to have an error string that prints out, as otherwise all the report says was "expect false to be true", and that's not useful

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ListingDetailsSteps.java
```diff
@@ -1,60 +1,71 @@
...
-    public void iOpenSEDAccordion() {
-        // to close certification criteria accordion that's open by default
-        ListingDetailsPage.certificationCriteriaAccordion(driver).click();
-
-        //to open SED accordion
-        ListingDetailsPage.sedAccordion(driver).click();
+    @Given("^I am on listing details page of listing with database ID \"(.*)\"$")
+    public void loadListingWithDbId(final String dbId) {
+        driver.get(url + "#/product/" + dbId);
+    }
```
andlar: Definitely want a "wait until" condition here, as the details page takes a long time to load

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ListingDetailsSteps.java
```diff
@@ -125,12 +120,73 @@ public void theCriteriaNumberShouldAlsoIncludeTheCriteriaTitle() {
...
+    public void theMandatoryDisclosuresUrlShouldBe(final String targetUrl) {
+        String actualText = ListingDetailsPage.disclosureUrl(driver).getText();
+        assertTrue(actualText.contains(targetUrl), "Expect " + targetUrl + " to be found in " + actualText);
+   }
+    /**
+     * Assert message when listing is not found.
+     */
+    @Then("^the page shows 'This listing does not exist' message$")
+    public void verifyMessageonPage() {
+        assertTrue(ListingDetailsPage.mainContent(driver).getText().contains("This listing does not exist"));
```
andlar: Would be helpful to have an error string on this assertTrue as well

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ListingDetailsSteps.java
```diff
@@ -125,12 +120,73 @@ public void theCriteriaNumberShouldAlsoIncludeTheCriteriaTitle() {
...
+        String actualString = ListingDetailsPage.ucdProcessText(driver).getText();
+        assertEquals(actualString, ucdText);
+    }
+    /**
+     * Asserts correct text in SED section.
+     * @param reportTitle expected text
+     */
+    @Then("^usability report text should read as \"([^\"]*)\"$")
+    public void usabilityReportTitleReadsAsFullUsabilityReport(final String reportTitle) {
+        assertTrue(ListingDetailsPage.usabilityReportTitle(driver).getText().contains(reportTitle));
```
andlar: Error string

src/test/resources/Features/OCD-1776-SEDTweaks.feature
```diff
@@ -1,33 +1,31 @@
 @Regression
-Feature: SED Display in product details
+Feature: OCD-1776 - SED Display in product details
   This test is to verify text changes and other display updates to SED section
 
-  Scenario: Verify SED details
-    Given I am on the Details page of Listing "8338"
+  Scenario Outline: Verify SED details
```
andlar: Why is this one an Outline? There's only one example, and I don't think it's likely we'd need to check multiple Listings to verify this Feature

src/test/resources/Features/OCD-1776-SEDTweaks.feature
```diff
@@ -1,33 +1,31 @@
...
-  Scenario: Verify SED Intended User Description header
-    Given I am on the Details page of Listing "8338"
-    When I open SED details accordion
-    Then intended user description header should show as 'Description of Intended Users'
-
-  Scenario: Verify SED download option and button title
-    Given I am on the Details page of Listing "8338"
-    When I open SED details accordion
-    Then the download button title should read as 'Download SED Details'
+  Scenario Outline: Verify usability report title update
```
andlar: Why is this an Outline too? It's not checking differences between Listings, right?

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
ashwinimore: No. I just wanted to move files only once in the beginning to copy any existing files in the folder, so we don't loose them before cleaning them from the directory and also so that the directory is empty before download process begins. It is not needed for later test runs.

andlar: Oh, I see what you mean. In that case, this is actually not a good thing. What if someone skips this test? Or they run out of order? I don't think there's a guarantee that all of the Scenarios run in the order they're in the file. Would it make more sense for the `this.movefiles();` command to be in a **Background:** section? That way it _will_ run once before all of the Scenarios, but only once.

Then maybe the `FileUtils.cleanDirectory(dir);` call could be in the `@Given I am on the CHPL Download page` just one time, but it'd run at the beginning of each Scenario.

src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
```diff
@@ -12,9 +12,36 @@
...
+    public static WebElement acbFiltersLink(final WebDriver driver) {
+        element = driver.findElement(By.xpath("//*[@id=\"filters\"]/div[2]/div[5]/ul/li[5]/st-list-multiple/ul/li[6]/label"));
+        return element;
+    }
+    /**
+     * Returns SLI Compliance ACB filter option.
+     * @param driver WebDriver
+     * @return SLI Compliance ACB filter option element
+     */
+    public static WebElement acbSLIFilter(final WebDriver driver) {
```
andlar, line 29: Didn't I add an ID to this link? Shouldn't the xpath be using that?

Also, if you want this to be an awesome element, it'd make sense to pass in the ACB name and use that, in conjunction with the IDs, to allow this PageObject call to reference any of the ACB filters

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -63,6 +43,10 @@ public ChplDownloadSteps() {
         if (StringUtils.isEmpty(url)) {
             url = "http://localhost:3000/";
         }
+        if (StringUtils.isEmpty(downloadPath)) {
+           downloadPath = "\\target";
```
andlar: My "target" directory has stuff, and likely will on any system that's running this test. I like using target as a starting point, but would suggest something like `\\target\\download-files` instead

andlar (changes requested):

andlar (approved):

src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
ashwinimore: I corrected the SLI element locator to use id, i had added xpath before id was added and forgot to update. I'll work on using ACB name in conjunction with the IDs to improve this test in next update.

PR #19, opened 2018-03-31
 fuzzymatch test fix, monitor download test update 

ashwinimore:
Review fix for fuzzymatch text test and monitor download file test update
@andlar 
@brianlindsy 
@tmy1313

### DIFF
--- a/src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
+++ b/src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
@@ -58,9 +58,10 @@ public class ChplDownloadSteps {
                 System.out.println(tempDirectory);
             } catch (final IOException e) {
                 // If temp directory creation failed, create new directory in target folder
+                // user.dir - User working directory, make new directories in user's working directory
                 File file = new File("target", "download-files");
                 file.mkdirs();
-                tempDirectory = System.getProperty("user.dir") + "\\target\\download-files";
+                tempDirectory = System.getProperty("user.dir") + File.separator + "target" + File.separator + "download-files";
                 System.out.println(tempDirectory);
             }
             downloadPath = tempDirectory;
--- a/src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
+++ b/src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
@@ -41,12 +41,13 @@ public class ChplDownloadSteps {
     /**
      * Constructor creates new driver.
      * Create a sub directory under default temp directory to set downloadPath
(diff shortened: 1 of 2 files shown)


ashwinimore:
I updated chromedriver and it's showing as conflict. Should I follow suggested steps below for resolution via command line?

andlar:
The chromedriver should be removed from the repository. It's not source code, and it's specific to each user's system, so it shouldn't be stored as part of the repository.

I was okay with it earlier when it didn't really matter, but as soon as it makes an impact (like this conflict) then it should be dealt with correctly.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -38,14 +40,24 @@
...
         }
         if (StringUtils.isEmpty(downloadPath)) {
-           downloadPath = "\\target\\download-files";
+            String tempDirectory;
+            try {
+              tempDirectory = Files.createTempDirectory("download-files").toString();
+              System.out.println(tempDirectory);
+            } catch (final IOException e) {
+              // Temp dir creation failed - use default
+              tempDirectory = null;
```
andlar: This doesn't "use default". If the directory creation fails it will use `null`, and then the directory will be `new File(null)`, which doesn't look right

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -51,13 +53,17 @@ public ChplDownloadSteps() {
...
+                // Print the path to the newly created directory
+                System.out.println(tempDirectory);
             } catch (final IOException e) {
-              // Temp dir creation failed - use default
-              tempDirectory = null;
+                // If temp directory creation failed, create new directory in target folder
+                File file = new File("target", "download-files");
+                file.mkdirs();
+                tempDirectory = System.getProperty("user.dir") + "\\target\\download-files";
+                System.out.println(tempDirectory);
```
andlar: Where does `file.mkdirs();` make the files/directories? Is it in `user.dir`? Are you sure? Is there always a property called user.dir? Also, the `+ "\\target\\download-files";` only works on Windows. On Linux the directory separator is a forward slash `/`.  You should use `File.pathSeparator` instead.

ashwinimore: yes, user.dir is where the directories should be created. When I run it locally, 'download-files' is created in target folder in my working directory. I am printing path to get exact location where directories will be created.

andlar (approved):

brianlindsy (approved):

tmy1313 (approved):

PR #23, opened 2018-04-06
Updates to batch withdraw feature to add missing steps

ashwinimore:
Please review updates:
@andlar 
@brianlindsy 
@tmy1313

### DIFF
--- a/src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
+++ b/src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
@@ -13,7 +13,6 @@ public class SearchPage {
 
     /**
      * Returns the acbFiltersLink.
-     *
      * @param driver webdriver
      * @return acbFilters Link
      */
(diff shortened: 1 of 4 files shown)


src/test/resources/Features/OCD-2134-batchWithdraw.feature
```diff
@@ -1,8 +1,24 @@
-Feature: Batch update Certification Status to 'Withdrawn by Developer' for a set of Cerner and Intermountain Healthcare listings
+Feature: OCD-2134 - Batch update of Certification Status to 'Withdrawn by Developer'
+		Batch update Certification Status to 'Withdrawn by Developer' for a set of Cerner and Intermountain Healthcare listings
+		
+Background: User is on search page and 'Withdrawn by Developer' filter option is checked
+	Given I am on CHPL search page
+	And I look at Certification Status options
+	And I apply "Withdrawn_by_Developer" filter
```
andlar: The Features files should read as if they were written by a human, and given to a non-programmer. To a non-programmer, the filter should be `Withdrawn By Developer` without the underscores.

src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
```diff
@@ -22,13 +21,22 @@ public static WebElement acbFiltersLink(final WebDriver driver) {
...
      * @param driver WebDriver
-     * @param acb is acb name
-     * @return SLI Compliance ACB filter option element
+     * @param selectfilter is desired filter name
+     * @return filter option element
      */
-    public static WebElement acbSLIFilter(final WebDriver driver, final String acb) {
-        element = driver.findElement(By.id("filter-list-" + acb));
+    public static WebElement filterOption(final WebDriver driver, final String selectfilter) {
+        element = driver.findElement(By.id("filter-list-" + selectfilter));
```
andlar: In this POM location, the passed in string should have spaces, and the POM should replace all of the spaces with underscores, something like `"filter-list-" + selectFilter.replaceAll(" ", "_")`

andlar (dismissed):
One small thing that shows up in two places. I don't think it's enough to reject the PR, so I'd be okay if you just merge it as is, but would prefer an update going forward at some point

andlar (approved):

brianlindsy (approved):

PR #26, opened 2018-04-20
OCD-1970 - Verify modification of G1/G2 values

ashwinimore:
@andlar 
@brianlindsy 
@tmy1313

### DIFF
--- a/src/test/java/gov/healthit/chpl/aqa/pageObjects/ListingDetailsPage.java
+++ b/src/test/java/gov/healthit/chpl/aqa/pageObjects/ListingDetailsPage.java
@@ -38,6 +38,7 @@ public final class ListingDetailsPage {
      * @param driver WebDriver
      * @param mName is measure Name
      * @return associatedCriteria element
+     * return null if criteria was not found in Associated Criteria column
      */
     public static WebElement associatedCriteria(final WebDriver driver, final String mName) {
         WebElement table = driver.findElement(By.xpath("//*[@id=\"panel-g1g2\"]/div/table/tbody"));
@@ -127,23 +128,14 @@ public final class ListingDetailsPage {
         return element;
     }
     /**
-     * Returns the element that holds G1 Measure Name for a criteria.
+     * Returns the element that holds G1/G2 Measure Name for a criteria.
      * @param driver WebDriver
      * @param number is criteria number
      * @return MeasureName element
+     * @param index is table row index for G1/G2
      */
-    public static WebElement g1MeasureName(final WebDriver driver, final String number) {
-        element = driver.findElement(By.xpath("//*[@id=\"criteria_" + number + "_details_header\"]/div/table/tbody/tr[2]/td[2]"));
-        return element;
-    }
-    /**
-     * Returns the element that holds G2 Measure Name for a criteria.
-     * @param driver WebDriver
-     * @param number is criteria number
-     * @return MeasureName element
-     */
-    public static WebElement g2MeasureName(final WebDriver driver, final String number) {
-        element = driver.findElement(By.xpath("//*[@id=\"criteria_" + number + "_details_header\"]/div/table/tbody/tr[3]/td[2]"));
+    public static WebElement gMeasureName(final WebDriver driver, final String number, final String index) {
+        element = driver.findElement(By.xpath("//*[@id=\"criteria_" + number + "_details_header\"]/div/table/tbody/tr[" + index + "]/td[2]"));
         return element;
     }
     /**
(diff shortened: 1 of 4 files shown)


src/test/resources/Features/OCD-1970-ModificationOfG1G2Values.feature
```diff
@@ -0,0 +1,51 @@
...
+    |8511|170.315 (b)(3)|Electronic Prescribing: Eligible Clinician|Required Test 1: ACI Transition Objective 2 Measure 1|
+    |8511|170.315 (b)(3)|Electronic Prescribing: Eligible Clinician Group|Required Test 9: ACI Objective 5 Measure 3|
+    |8511|170.315 (g)(8)|View, Download, or Transmit (VDT):  Eligible Clinician Individual (TIN/NPI)|Required Test 4: ACI Objective 4 Measure 1|     
+	|8511|170.315 (g)(8)|View, Download, or Transmit (VDT):  Eligible Clinician Group|Required Test 4: ACI Transition Objective 3 Measure 2|
+	|8511|170.315 (g)(8)|View, Download, or Transmit (VDT):  Eligible Provider Individual|Required Test 4: Stage 2 Objective 8 Measure 2|
+	|8511|170.315 (g)(9)|View, Download, or Transmit (VDT):  Eligible Provider|Required Test 4: Stage 3 Objective 6 Measure 1|
+	|8511|170.315 (g)(9)|View, Download, or Transmit (VDT):  Eligible Provider|Required Test 4: Stage 2 Objective 8 Measure 2|
+	|8511|170.315 (g)(9)|View, Download, or Transmit (VDT):  Eligible Provider|Required Test 2: Stage 3 Objective 5 Measure 1|	
+	
+Scenario Outline: Mapping legacy values for G1
```
andlar: I could see this Scenario and the previous one being combined, using a "Then" step of "Measure Successfully Tested for `<gOption>` should display..."

src/test/resources/Features/OCD-1970-ModificationOfG1G2Values.feature
```diff
@@ -0,0 +1,51 @@
+@Regression
+Feature: OCD-1970 - Addition / modification of available G1/G2 values
+
+Scenario Outline: Mapping legacy values for G2
+	Given I am on listing details page of listing with database ID "<DB_ID>"
+	When I open details for criteria "<CRITERIA>"
+	Then Measure Successfully Tested for G2 should display "<Measure_Name>" for "<CRITERIA>" 
+	And When I look at  Successfully Tested G1G2 Measures
```
andlar: Make this a second "When-Then" set. So the whole thing would be
```Given
When
Then
When
Then
```

I wasn't sure if that was okay, but https://stackoverflow.com/questions/12060011/is-it-acceptable-to-write-a-given-when-then-when-then-test-in-gherkin says it is

src/test/java/gov/healthit/chpl/aqa/pageObjects/ListingDetailsPage.java
```diff
@@ -152,6 +191,23 @@ public static WebElement nosedData(final WebDriver driver) {
         element = driver.findElement(By.xpath("//*[@id=\"panel-sed\"]/div/p"));
         return element;
     }
+    /**
+     * Returns the element that holds Required Text for a Measure Name in Successfully Tested G1/G2 Measures section .
+     * @param driver WebDriver
+     * @param mName is measure Name
+     * @return requiredText element
```
andlar, line 190: Should mention something in the javadoc for when "null" could be returned

src/test/java/gov/healthit/chpl/aqa/pageObjects/ListingDetailsPage.java
```diff
@@ -107,6 +126,26 @@ public static WebElement g1g2Accordion(final WebDriver driver) {
...
+        return element;
+    }
+    /**
+     * Returns the element that holds G2 Measure Name for a criteria.
+     * @param driver WebDriver
+     * @param number is criteria number
+     * @return MeasureName element
+     */
+    public static WebElement g2MeasureName(final WebDriver driver, final String number) {
+        element = driver.findElement(By.xpath("//*[@id=\"criteria_" + number + "_details_header\"]/div/table/tbody/tr[3]/td[2]"));
```
andlar: This will only find the right row if the Listing has the possibility of both GAP and a Standard. Do all of the G1/G2 measures have both of those? If not we should probably add an ID to the rows of the criteria details section

andlar (changes requested):
I've only just realized. The best way to review these PRs is to start at the bottom, and go through the Feature, then the Steps, then the Page Objects. Probably best to look at comments that way too :)

Edited to add: Oh, weird! The comments are put in the review in the order I wrote them, so the comments can be read top down, even though I wrote them bottom up.

tmy1313 (dismissed):

brianlindsy (dismissed):

src/test/java/gov/healthit/chpl/aqa/pageObjects/ListingDetailsPage.java
ashwinimore: Adding ids to rows is good idea. For this test, the test data i have used for existing listings works well for selected criteria in the listings.

andlar (approved):

PR #28, opened 2018-04-27
OCD-2102 - Allow surveillance search for listings with CHPL ID version

ashwinimore:
Added new tests - login and surveillance search with CHPL ID version.
@andlar 
@brianlindsy 
@tmy1313

### DIFF
--- a/src/test/java/gov/healthit/chpl/aqa/pageObjects/DpManagementPage.java
+++ b/src/test/java/gov/healthit/chpl/aqa/pageObjects/DpManagementPage.java
@@ -1,6 +1,5 @@
 package gov.healthit.chpl.aqa.pageObjects;
 
-
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
@@ -22,6 +21,7 @@ public class DpManagementPage {
         element = driver.findElement(By.id("listing-file-upload"));
         return element;
     }
+
     /**
      * Returns choose file button for bulk surveillance upload.
      * @param driver WebDriver
@@ -31,6 +31,7 @@ public class DpManagementPage {
         element = driver.findElement(By.id("surveillance-file-upload"));
         return element;
     }
+
     /**
      * Get the CHPL Product Number value in surveillance search results.
      * @param driver WebDriver
@@ -41,6 +42,7 @@ public class DpManagementPage {
         element = driver.findElement(By.xpath("//*[@id=\"main-content\"]/section/div/div/div/div[3]/table/tbody/tr[1]/td"));
         return element;
     }
+
     /**
      * Returns Developer & Product Management link.
      * @param driver WebDriver
@@ -50,6 +52,7 @@ public class DpManagementPage {
         element = driver.findElement(By.linkText("Developer & Product Management"));
         return element;
     }
+
     /**
      * Returns Manage Surveillance Activity link.
      * @param driver WebDriver
@@ -59,6 +62,7 @@ public class DpManagementPage {
         element = driver.findElement(By.linkText("Manage Surveillance Activity"));
         return element;
     }
+
     /**
      * Returns Confirm Pending Products link.
      * @param driver WebDriver
@@ -68,6 +72,7 @@ public class DpManagementPage {
         element = driver.findElement(By.linkText("Confirm Pending Product"));
         return element;
     }
+
     /**
      * Returns Pending Listings table in Confirm Pending Listings section.
      * @param driver WebDriver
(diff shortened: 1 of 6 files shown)


src/test/java/gov/healthit/chpl/aqa/pageObjects/DpManagementPage.java
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Returns choose file button for new listing upload.
+     * @param driver WebDriver
+     * @return button element
+     */
+    public static WebElement chooseFileForNewListingUploadButton(final WebDriver driver) {
+        element = driver.findElement(By.id("listing-file-upload"));
+        return element;
+    }
+    /**
```
andlar: It's not a huge deal. For that matter, it's probably super picky, but I like to have a blank line between the previous method and the javadoc for the next one.

The best part of this comment is that, if this is the level of comment I'm reduced to, we're in a really good place as far as code goes... :)

src/test/resources/Features/login.feature
```diff
@@ -0,0 +1,9 @@
+Feature: User login
+
+Scenario: User logs in as ACB Admin
+	Given I am on CHPL admin page	
+	When I fill in username
+	And I fill in Password
+	When I click Log in button
```
andlar: Should be an "And", not a "When". These scenarios should make sense to a customer as "English".

src/test/resources/Features/OCD-2102-SearchSurveillanceListingsWithChplIDVersion.feature
```diff
@@ -0,0 +1,17 @@
+@Regression
+Feature: OCD-2102 - Allow surveillance search for listings with CHPL ID version
+
+Background: User is logged in and navigates to Manage Surveillnace Activity section
+	Given I am on CHPL admin page	
+	When I fill in username
+	And I fill in Password
+	When I click Log in button
+	Then I navigate to Manage Surveillnace Activity section
```
andlar: I don't like this as a "background" thing. Part of it might be that I don't like our current admin nav section, but even without that, the GWAWT should probably be something more like GWAAT.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/LoginSteps.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    @Then("^I should be logged in to CHPL as ACB Admin$")
+    public void verifyLoginWasSuccessful() {
+        String actualString = LoginPage.welcomeText(driver).getText();
+        assertTrue(actualString.contains("Welcome"));
+    }
+    /**
+     * User logout.
+     */
+    @Then("^I log out$")
```
andlar: This step isn't in any Feature; is it needed?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/LoginSteps.java
```diff
@@ -0,0 +1,90 @@
...
+        driver = Hooks.getDriver();
+        if (StringUtils.isEmpty(url)) {
+            url = "http://localhost:3000/";
+           }
+
+        if (StringUtils.isEmpty(username)) {
+            throw new IllegalArgumentException("Missing value for username!");
+           }
+        if (StringUtils.isEmpty(password)) {
+            throw new IllegalArgumentException("Missing value for password!");
```
andlar, line 39: Yes!

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/DpManagementSteps.java
```diff
@@ -0,0 +1,57 @@
...
+import cucumber.api.java.en.When;
+import gov.healthit.chpl.aqa.pageObjects.DpManagementPage;
+
+/**
+ * Class DpManagementSteps definition.
+ */
+
+public class DpManagementSteps {
+    private WebDriver driver;
+    private static final int TIMEOUT = 30;
```
andlar: Is TIMEOUT used here? If not, it shouldn't be declared. If it should be used, then I don't see it...

andlar (changes requested):

src/test/resources/Features/OCD-2102-SearchSurveillanceListingsWithChplIDVersion.feature
ashwinimore: I added it as background in case there was a scenario at later point to validate other fields in results. That said, it may be another feature altogether. I'll remove background and add it to scenario.

andlar (approved):

PR #31, opened 2018-05-04
OCD-2049 - Verify 'Eye' on listing details page works and displays product history

ashwinimore:
Added a test to verify that product history eye works.
@andlar 
@brianlindsy 
@tmy1313 
@kekey1


andlar (approved):
The only thing I could think to add would be for a "null" result. As in, using one of the Listings that doesn't have any history, an "expect" that it says (whatever it says when there's not history).

No requirement for it now, but if you have a chance, or time, to do it, wouldn't hurt.

ashwinimore:
I'll add a test for it in next update, as  I am planning to add one more test to it for legacy listings that show a different message/text in history.

PR #32, opened 2018-05-09
OCD-2213-Tests to verify missing listing reappears in search results

ashwinimore:


src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
```diff
@@ -154,4 +154,24 @@ public static WebElement searchField(final WebDriver driver) {
...
+        element = driver.findElement(By.xpath("//*[@id=\"filters\"]/div[2]/div[1]/ul/st-list-multiple/ul/li[1]/a[1]"));
+        return element;
+    }
+
+    /**
+     * Returns item count text on search results.
+     * @param driver WebDriver
+     * @return text element
+     */
+    public static WebElement totalItemCount(final WebDriver driver) {
```
andlar: In my version of the code there's already a WebElement called "resultCount" that has the same xpath.

andlar (changes requested):

andlar (approved):
We might want to chat at some point about organization of page object elements to avoid this kind of thing in the future. Maybe placed in the file in the order the items appear on the page? Worth thinking about

kekey1 (approved):

brianlindsy (approved):

ashwinimore:
We talked about it in the beginning and came up with organizing in alphabetical order, it used to be by appearance of element on page. It was simply my bad that i missed it when i scrolled past it to check for existing elements. That said, I agree that with consolidated steps and page objects growing, it's definitely worth looking into alternative approach. I'll do some research on best practices to follow with page objects.

PR #33, opened 2018-05-10
Tests to verify missing listings added to the CHPL

ashwinimore:


andlar (approved):
Not necessarily now, but I would change line one in "addMissingListingstoCHPL.feature" to be "Feature: Verify Listings exist on CHPL, and maybe rename the file in the same way. It's a slightly more generic phrase, and will cover more tickets later

ashwinimore:
I'll add feature title change in next update.

PR #39, opened 2018-06-18
Fix broken tests in G1/G2 update and surveillance search features

ashwinimore:


andlar (approved):
Sorry about the id change. Generally speaking, the goal is to move all ids to the lowercase-dash format, though it's likely to be a long term project. Will try to let you know when I do one of these changes that you depend on

PR #42, opened 2018-07-10
OCD-2006-Upload fake listings regularly to test upload/confirm workflow

ashwinimore:


andlar (approved):
I don't see how this has a new CHPL ID for each day, but if you think it works then I'm happy to see it happen :)

kekey1 (approved):

ashwinimore:
product code is being set to ddmm everyday for that specific day, which makes (will make) CHPL ID unique for daily uploads.

PR #44, opened 2018-07-12
OCD-2357 - Verify Batch Withdraw update - Cerner (ICSA final set of l…

ashwinimore:
…istings)


andlar (changes requested):
Are "OCD-2134-batchWithdraw.feature" and "batchWithdraw.feature" the same file? If so, shouldn't there only be one of them?

Secondly, there are some weird indentation things going on in the feature files, and I think it's because your editor is set up to use tabs sometimes, and spaces others. It's not a big issue, but it would be nice to have consistency. For the java code in CHPL we're set at using four spaces for each indent, and that would make sense for the features files too

ashwinimore:
I edited filename from OCD-2134-batchWithdraw.feature to have batchWithdraw.feature as new name. I no longer have OCD-2134-batchWithdraw.feature on my local.

andlar (approved):

kekey1 (approved):

PR #45, opened 2018-07-19
update fake listings upload script to run successfully on Jenkins

ashwinimore:
moved upload files to test/resources.


pom.xml
```diff
@@ -122,7 +122,12 @@
 		    <groupId>org.apache.commons</groupId>
 		    <artifactId>commons-csv</artifactId>
 		    <version>1.5</version>
-			</dependency>			
+			</dependency>
+			<dependency>
+	      <groupId>com.google.guava</groupId>
```
andlar: What's this dependency for? I just ran my version of the tests and I think they ran successfully. Is it possible adding this dependency is your problem?

ashwinimore: I added to get something working but it's not needed. I am going to remove it.

andlar (approved):

PR #50, opened 2018-08-29
Update upload listing tests to take screenshot at failure

ashwinimore:


kekey1 (approved):

src/test/java/gov/healthit/chpl/aqa/pageObjects/DpManagementPage.java
```diff
@@ -47,6 +51,25 @@ public static WebElement confirmPendingProductsLink(final WebDriver driver) {
...
+     * @param testChplId is chpl id of listing to confirm
+     * @return the inspect button element
+     * @throws Exception if there is an exception
+     */
+    public static WebElement inspectButtonForUploadedListing(final WebDriver driver, final String testChplId) {
+        WebElement table = driver.findElement(By.id("pending-listings-table"));
+        ArrayList<WebElement> rows = (ArrayList<WebElement>) table.findElement(By.tagName("tbody")).findElements(By.tagName("tr"));
+        for (WebElement row : rows) {
+            ArrayList<WebElement> cols = (ArrayList<WebElement>) row.findElements(By.tagName("td"));
+            if (cols.get(0).getText().equalsIgnoreCase(testChplId)) {
```
andlar, line 66: I like this, but think you should add an `ID_COLUMN` like the `ACTION_COLUMN`. I know checkstyle doesn't give us "magic number" errors for 0, 1, and maybe 2, but that's more to be nice to us then because they're not magic numbers.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/UploadListingsRegularlySteps.java
```diff
@@ -144,8 +136,19 @@ public void confirmUploadedListing(final String edition, final String testChplId
...
+        } catch (Exception e) {
+            takeScreenshot();
+        }
+    }
 
+    /**
+     * Take a screenshot.
+     * @throws Exception if there is an exception
+     */
+    public void takeScreenshot() throws Exception {
```
andlar, line 149: Good start. We'll likely want this function to exist in all of the Step files, so would recommend moving it to a "BaseStep.java" class and have at least this step file extend that base file.

Also, should do something to make the filename more unique; if more than one screenshot is taken during a run they'll overwrite each other. Could append the timestamp, with something like `filePath + File.separator = "failed-test-" + (new Date()).getTime() + ".png"`

andlar (commented):
I'm not against merging this in; the suggestions below are not required. Up to you

src/test/java/gov/healthit/chpl/aqa/pageObjects/DpManagementPage.java
ashwinimore: I'll update that.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/UploadListingsRegularlySteps.java
ashwinimore: I have it in my 'TO-DO' list :)

PR #70, opened 2018-11-02
OCD-2505-Verify Implementation Note update for certified_products end…

ashwinimore:
…point

### DIFF
--- a/src/test/java/gov/healthit/chpl/aqa/pageObjects/ChplAPIPage.java
+++ b/src/test/java/gov/healthit/chpl/aqa/pageObjects/ChplAPIPage.java
@@ -11,17 +11,7 @@ public final class ChplAPIPage {
     private ChplAPIPage() {}
     private static WebElement element = null;
 
-    /**
-     * Returns element holding main content.
-     * @param driver WebDriver
-     * @return element holding main content
-     */
-    public static WebElement mainContent(final WebDriver driver) {
-        element = driver.findElement(By.id("main-content"));
-        return element;
-    }
-
-    /**
+     /**
      * Controller link.
      * @param driver WebDriver
      * @return controller link element
@@ -29,16 +19,6 @@ public final class ChplAPIPage {
      */
     public static WebElement controllerLink(final WebDriver driver, final String controller) {
         element = driver.findElement(By.xpath("//*[@id=\"" + controller + "*\"]/li/h4/a"));
-        return element;
-    }
-
-    /**
-     * Certified Product endpoint link.
-     * @param driver WebDriver
-     * @return link element
-     */
-    public static WebElement certifiedProductsEndpoint(final WebDriver driver) {
-        element = driver.findElement(By.xpath("//*[@id=\"getCertifiedProductsByVersionUsingGET\"]/a/div/span[3]"));
         return element;
     }
 
@@ -51,4 +31,16 @@ public final class ChplAPIPage {
         element = driver.findElement(By.xpath("//*[@id=\"main-content\"]/div[3]/div/div[2]/div/ul/li[19]/ul/li[1]/div[2]/div/p"));
         return element;
     }
+
+    /**
+     * Endpoint link.
+     * @param driver WebDriver
+     * @return link element
+     * @param endpoint is given endpoint to view details
+     */
+    public static WebElement endpointLink(final WebDriver driver, final String endpoint) {
+        element = driver.findElement(By.xpath("//*[@id=\"" + endpoint + "\"]/a/div/span[1]"));
+        return element;
+    }
+
 }
--- a/src/test/java/gov/healthit/chpl/aqa/pageObjects/ChplAPIPage.java
+++ b/src/test/java/gov/healthit/chpl/aqa/pageObjects/ChplAPIPage.java
@@ -7,9 +7,9 @@ import org.openqa.selenium.WebElement;
 /**
  * Class ChplAPIPage definition.
  */
-public final class ChplAPIPage {
+public final class ChplAPIPage extends BasePage {
+
     private ChplAPIPage() {}
-    private static WebElement element = null;
 
      /**
      * Controller link.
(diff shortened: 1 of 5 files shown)


src/test/java/gov/healthit/chpl/aqa/pageObjects/ChplAPIPage.java
```diff
@@ -0,0 +1,54 @@
...
+public final class ChplAPIPage {
+    private ChplAPIPage() {}
+    private static WebElement element = null;
+
+    /**
+     * Returns element holding main content.
+     * @param driver WebDriver
+     * @return element holding main content
+     */
+    public static WebElement mainContent(final WebDriver driver) {
```
andlar: I wonder if we could do something where we have a "BasePage" class that has this element in it, along with the navigation ones, and have all of our specific pages extend from that. Then all pages would have a reference to the main-content as well as navigation things, and we'd have the code in only one place

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPIDocumentationSteps.java
```diff
@@ -0,0 +1,75 @@
...
+import cucumber.api.java.en.And;
+import cucumber.api.java.en.Given;
+import cucumber.api.java.en.Then;
+import cucumber.api.java.en.When;
+import gov.healthit.chpl.aqa.pageObjects.ChplAPIPage;
+
+/**
+ * Class ChplAPIDocumentationSteps definition.
+ */
+public class ChplAPIDocumentationSteps {
```
andlar: Take a look at "ChplDownloadSteps" and "ManageDevelopersAndProductsSteps" for examples of steps pages that extend the BaseSteps class. By extending BaseSteps you don't have to manage the webdriver, url, timeouts, etc.

src/test/resources/Features/APIDocumentationUpdate.feature
```diff
@@ -0,0 +1,12 @@
+Feature: API Documentation updates
+         OCD-2505 - Update Implementation Note for /certified_products endpoint
+
+Scenario Outline: Verify that implementation note for /certified_products endpoint is updated with new text
+    Given I am on CHPL API page
+    When I click on "<controller_name>" to see operations list
+    And I click on Certified Products endpoint link to view details
```
andlar: Either the "When" and "Then" steps shouldn't be parameterized, or the "And" step should. There's no way to do a second Example without changing the "And".

The "solves the feature" option would be to not parameterize either of them, while the "prepares for the future" option would be to do it for them all. If you _do_ do it for them all, I'd recommend putting an Example in that should have been successful even before the story started

andlar (commented):
I have some comments, but don't think they're required changes. If you want to address them that's great, if not consider this an approval

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/BaseSteps.java
```diff
@@ -63,4 +66,15 @@ public WebDriverWait getWait() {
     public void setWait(final WebDriverWait wait) {
         this.wait = wait;
     }
+
+    /**
+     * Returns element holding main content.
+     * @param driver WebDriver
+     * @return element holding main content
+     */
+    public static WebElement mainContent(final WebDriver driver) {
```
andlar: I like this, but it's in the wrong file. It should be in a new "BasePage.java" file in the pages directory, not in the steps file

andlar (changes requested):
Love the changes. Just need to move the Page part into a BasePage class and we're set. Might add another Example to the Feature just for completeness, but that's not required.

src/test/java/gov/healthit/chpl/aqa/pageObjects/ChplAPIPage.java
```diff
@@ -0,0 +1,46 @@
+package gov.healthit.chpl.aqa.pageObjects;
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+/**
+ * Class ChplAPIPage definition.
+ */
+public final class ChplAPIPage {
```
andlar:
```suggestion
public final class ChplAPIPage extends BasePage {
```

This class needs to extend the BasePage, then any calling step can do `ChplApiPage.mainContent` and get that element

src/test/java/gov/healthit/chpl/aqa/pageObjects/ChplAPIPage.java
```diff
@@ -0,0 +1,46 @@
...
+    private ChplAPIPage() {}
+    private static WebElement element = null;
```
andlar:
```suggestion
```
This isn't needed

src/test/java/gov/healthit/chpl/aqa/pageObjects/ChplAPIPage.java
```diff
@@ -0,0 +1,46 @@
...
+
+     /**
+     * Controller link.
+     * @param driver WebDriver
+     * @return controller link element
+     * @param controller desired controller to open
+     */
+    public static WebElement controllerLink(final WebDriver driver, final String controller) {
+        element = driver.findElement(By.xpath("//*[@id=\"" + controller + "*\"]/li/h4/a"));
```
andlar:
```suggestion
        return driver.findElement(By.xpath("//*[@id=\"" + controller + "*\"]/li/h4/a"));
```
Can delete the next line, and do the same changes for the other PageObject elements

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPIDocumentationSteps.java
```diff
@@ -0,0 +1,68 @@
...
+       }
+
+    /**
+     * Get user to the API Documentation page.
+     */
+    @Given("^I am on CHPL API page$")
+    public void userLoadsAPIPage() {
+        getDriver().get(getUrl() + "#/resources/chpl_api");
+        WebDriverWait wait = new WebDriverWait(getDriver(), TIMEOUT);
+        wait.until(ExpectedConditions.visibilityOf(BasePage.mainContent(getDriver())));
```
andlar:
```suggestion
        wait.until(ExpectedConditions.visibilityOf(ChplAPIPage.mainContent(getDriver())));
```
With ChplApiPage extending BasePage, you can treat the ChplApiPage as having all of the BasePage content

andlar (changes requested):

andlar (approved):

kekey1 (approved):

PR #75, opened 2018-11-24
AQA OCD-2262 - Remove API calls that were deprecated

iyerashwin14:


src/test/resources/Features/OCD - 2262-RemoveDeprecatedApiPostCallsOfPutOrDelete.feature
```diff
@@ -0,0 +1,50 @@
+Feature: OCD-2262 - Remove API POST calls that were deprecated in favor of PUT/DELETE 
+    This test checks whether deprecated calls are removed from CHPL API webpage
+
+@Regression
+Scenario: CHPL API page loads and shows correct title 
+    Given I am on CHPL API page 
+    Then CHPL API page title should be "CHPL API" 
+
+Scenario Outline:
```
andlar, line 10: When is this outline run? It's not tagged with anything, and looks very similar to the Acceptance scenario below

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ExcelUtil.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+public final class ExcelUtil {
+    private static XSSFSheet excelWSheet;
+    private static XSSFWorkbook excelWBook;
+    private static XSSFCell cell;
+
+    /**
+     * String path points to the location of the ChplApi.xlsx, includes all the APIs based on the categories.
+     */
+    private static String path = System.getProperty("user.dir") + "\\src\\test\\resources\\ChplApi.xlsx";
```
andlar: Should use the `File.separator` thing to do file locations, as it's platform agnostic

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +34,56 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+     * Assert that contents from webpage matches the contents of excel.
+     * @param str as sheet name in the excel
+     */
+    @Then("^list of \"([^\"]*)\" operations should be displayed without deprecated calls$")
+    public void listOfOperationsShouldBeDisplayed(final String str) {
+        String sheetName = null;
+        WebDriver driver = getDriver();
+        // Get element from the webpage into the list
+        List<WebElement> listWebElements = ChplAPIPage.controllerElementList(driver);
+        if (str.equalsIgnoreCase(CORR_ACT_PLN_CTRL)) {
```
andlar: The corrective-action-plan controller is entirely deprecated, and probably doesn't need to be here at all

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +34,56 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+        // Get text of the Web Elements list
+        for (WebElement wElement : listWebElements) {
+            apiListFromPage.add(wElement.getText());
+        }
+        List<String> apiLstExl = new ArrayList<>();
+        // Getting api list from the excel into the list
+        apiLstExl = ExcelUtil.getAllCellData();
+        // Checking if the api from the webpage present in the excel
+        for (String api : apiListFromPage) {
+            assertEquals(apiLstExl.contains(api), true);
```
andlar: Doesn't this check that the API calls you have in the Excel sheet exist on the page? Where's the check that the ones that used to be deprecated don't exist at all?

Also, instead of `assertEquals(something, true)` it'd make more sense if you used `assertTrue(something)`

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +34,56 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+        Assert.assertEquals(chplAPITitle, title);
+    }
+
+    /**
+     * Click all the controller elements link.
+     * @param arg1 as controller elements
+     */
+    @When("^I click on \"([^\"]*)\" link$")
+    public void iClickOnLink(final String arg1) {
+        String xPath = "//*[text()='" + arg1 + "']";
```
andlar: Steps shouldn't be using xPaths. Only Page Objects should have xPath stuff. Steps could pass arguments to pages if there are multiple types of things, like in the API lists

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +34,56 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+        String chplAPITitle = getDriver().getTitle();
+        Assert.assertEquals(chplAPITitle, title);
+    }
+
+    /**
+     * Click all the controller elements link.
+     * @param arg1 as controller elements
+     */
+    @When("^I click on \"([^\"]*)\" link$")
+    public void iClickOnLink(final String arg1) {
```
andlar: It's good coding practice to use meaningful names for variables

andlar (changes requested):
I'm not sure how I feel about using an Excel file as part of our work. Excel is a proprietary format, and though it looks like Java can handle it, I'd much rather have csvs, as they're strictly text, which means changes in them can be tracked in git. Also, for what you're using them, they could be done as a set of examples in the scenario, where the first column is the controller, and the second is the API endpoint itself

src/test/resources/Features/OCD - 2262-RemoveDeprecatedApiPostCallsOfPutOrDelete.feature
ashwinimore: Add @Regression tag to this scenario

ashwinimore (changes requested):
Agree with Andrew's review comments

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
iyerashwin14: According to my current logic, we are taking all the APIs from the Webpage related to each controller in a list.  Later, reading all the APIs from the excel sheet considering that excel sheet will have only the non-deprecated APIs. Now, we are iterating each API from the webpage into the API list and then reading from the excel sheet. If there is any deprecated API present in the webpage that's not present in the excel sheet, assert will fail.

In this way, we will make sure that till the time all deprecated APIs are not removed from the webpage the test will keep on failing. **Assuming that excel data contains only the non-deprecated APIs.**

Now,
As per my understanding, the excel or CSV should contain only Deprecated APIs and check if these endpoints are present on the webpage or not.
If they are present, we will consider it as failed because webpage is containing the deprecated APIs and if not, it should pass. 

Please let me know if I'm correct

andlar: I see what you mean. I hadn't understood the logic that you were using. I think the methodology you have here is now fine, though I still don't like using the Excel file.

Could you do it with scenarios, where one value is the controller, and the second value is a delimited list of API calls? Then they could be stored directly in the Feature file, and easily modified if/when we change API endpoints again

Something like:
```
+--------------------+--------------------------------------------------------------------------------+
|Controller          |API Endpoints                                                                   |
+--------------------+--------------------------------------------------------------------------------+
|Certified Product   |Upload a new certified product; confirm a new certified product; etc...         |
+--------------------+--------------------------------------------------------------------------------+
|Developer           |Create a developer; get a developer; ...                                        |
+--------------------+--------------------------------------------------------------------------------+
|Etc...              |Etc...                                                                          |
+--------------------+--------------------------------------------------------------------------------+
```

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +33,46 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+        // Creating a list to hold all the API listed on the webpage by swagger
+        List<String> apiListFromPage = new ArrayList<>();
+        // Get text of the Web Elements list
+        for (WebElement wElement : listWebElements) {
+            apiListFromPage.add(wElement.getText());
+        }
+        System.out.println("apiEndpoints" + apiEndpoints);
+        // Checking if the API from the webpage is present in the example API endpoints
+        for (String api : apiListFromPage) {
+            assertTrue(apiEndpoints.contains(api));
```
andlar: This is an excellent change. The only thing I see now is that  your check for the API call on the page to be in the Examples data could fail if someone edited an API call and removed words from the end. For example, if the page had the API call "Create a new trigger" that would be found in the string related to the scheduler-controller, even though it's not the complete API call as defined in that data. Would recommend taking the apiEndpoints value and splitting it into an array or List, then using some List.contains(api) call to check to see that it matches exactly.

Also consider using something other than a comma for delineating the API calls; it's possible for an API description to have a comma in it. Could probably use multiple characters if that'd help. I was going to suggest a pipe, but that doesn't work with the feature definition. Maybe `*`, or `--`, or even something silly like `xxx`

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
tmy1313: Do you need the 'apiListFromPage' list?
Could you do the following and eliminate the need to 2 loops?
```
for (WebElement element : listWebElements) {
    assertTrue(element.getText().contains(api));
}
```

src/test/resources/Features/OCD - 2262-RemoveDeprecatedApiPostCallsOfPutOrDelete.feature
```diff
@@ -0,0 +1,48 @@
+Feature: OCD-2262 - Remove API POST calls that were deprecated in favor of PUT/DELETE 
+    This test checks whether deprecated calls are removed from CHPL API webpage
+
+@Regression
+Scenario: CHPL API page loads and shows correct title 
+    Given I am on CHPL API page 
+    Then CHPL API page title should be "CHPL API"
```
andlar: Please generalize this step

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +34,51 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+        List<String> apiListFromPage = new ArrayList<>();
+        // Get text of the Web Elements list
+        for (WebElement wElement : listWebElements) {
+            apiListFromPage.add(wElement.getText());
+        }
+        // Splitting string into tokens of APIs separated by "--" and putting into a list
+        List<String> apiListFromEndpoints = Arrays.asList(apiEndpoints.split("--"));
+        // Checking if the API from the webpage is present in the API Endpoints
+        for (String apiPage : apiListFromPage) {
+            assertTrue(apiListFromEndpoints.contains(apiPage), "API is deprecated/newly added api not found in the API Endpoints:-" + apiPage);
```
andlar: Move this into the loop at line 71

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/BaseSteps.java
```diff
@@ -19,4 +21,14 @@ public void iclickCHPLResourcesTopNavigationLink() {
         WebDriverWait wait = new WebDriverWait(driver, TIMEOUT);
         wait.until(ExpectedConditions.visibilityOf(BasePage.chplResourcesDropdown(driver))).click();
     }
+
+    /**
+     * Assert that title should be CHPL API.
+     * @param title expected as CHPL API
+     */
+    @Then("^Page title should be \"([^\"]*)\"$")
+    public void chplApiPageTitleShouldBe(final String title) {
```
andlar: The name of this method is not generic. Same with the variable `chplAPITitle`

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -2,9 +2,19 @@
...
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.testng.Assert;
+import static org.testng.Assert.assertTrue;
```
andlar: If you do `Ctrl-Shift-O` in Eclipse it will automatically re-order your imports. I mention it because that puts static imports together up at the top, and helps make the imports consistent across files

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +34,50 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
         assertEquals(actualText, text);
         assertEquals(actualURL, urlLink);
     }
+
+    /**
+     * Assert that title should be CHPL API.
+     * @param title expected as CHPL API
+     */
+    @Then("^CHPL API page title should be \"([^\"]*)\"$")
+    public void chplApiPageTitleShouldBe(final String title) {
```
andlar: This method is now not required, right?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +34,50 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+     * Assert that APIs from webpage is present in the example API Endpoints.
+     * @param apiEndpoints as the details of controller
+     */
+    @Then("^list of \"(.*)\" operations should be displayed without deprecated calls$")
+    public void listOfOperationsShouldBeDisplayed(final String apiEndpoints) {
+        WebDriver driver = getDriver();
+        // Get element from the webpage into the list
+        List<WebElement> listWebElements = ChplAPIPage.controllerElementList(driver);
+        // Creating a list to hold all the API listed on the webpage by swagger
+        List<String> apiListFromPage = new ArrayList<>();
```
andlar, line 54: This variable is filled, but then never read

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +34,50 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+    }
+
+    /**
+     * Assert that APIs from webpage is present in the example API Endpoints.
+     * @param apiEndpoints as the details of controller
+     */
+    @Then("^list of \"(.*)\" operations should be displayed without deprecated calls$")
+    public void listOfOperationsShouldBeDisplayed(final String apiEndpoints) {
+        WebDriver driver = getDriver();
+        // Get element from the webpage into the list
```
andlar: I don't think you need all of these comments, but if you're going to have them, they ought to be accurate and grammatically correct. In this case it's not `Get element from...` but instead `Get elements from...`, though the `...into the list` is weird. Maybe `Load elements from the page into a list`?

Honestly, and this might just be me, trying to read code where every other line is a comment is jarring. It's like I have to switch back and forth from English to Java constantly, and that's disconcerting

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +34,50 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+        List<WebElement> listWebElements = ChplAPIPage.controllerElementList(driver);
+        // Creating a list to hold all the API listed on the webpage by swagger
+        List<String> apiListFromPage = new ArrayList<>();
+        //Splitting string into tokens of APIs separated by "--" and putting into a list
+        List<String> apiListFromEndpoints = Arrays.asList(apiEndpoints.split("--"));
+        // Checking if the API from the webpage is present in the API Endpoints
+        for (WebElement wElement : listWebElements) {
```
andlar: `wElement` is a variable name that doesn't tell me anything about what's in the variable. This is one of the specific API elements, right? Why not 'apiLink'?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplAPISteps.java
```diff
@@ -24,4 +34,50 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+        // Get text of the Web Elements list
+            String apiName = wElement.getText();
+            apiListFromPage.add(apiName);
+            assertTrue(apiListFromEndpoints.contains(apiName), "API is deprecated/newly added api not found in the dataset:-" + apiName);
```
andlar: Don't need either of the two previous lines, and this one could just be
```suggestion
            assertTrue(apiListFromEndpoints.contains(apiName), "Unknown API found: " + apiName);
```

src/test/resources/Features/OCD - 2262-RemoveDeprecatedApiPostCallsOfPutOrDelete.feature
```diff
@@ -0,0 +1,49 @@
+Feature: OCD-2262 - Remove API POST calls that were deprecated in favor of PUT/DELETE 
+    This test checks whether deprecated calls are removed from CHPL API webpage
+
+@Regression
+Scenario: CHPL API page loads and shows correct title 
+    Given I am on CHPL API page 
+    Then Page title should be "CHPL API"
```
andlar: I'd prefer these steps be written as close to English as possible. So:
```
Given I am on the CHPL API page
Then the page title should be "CHPL API"
```
If you read the steps out loud it should sound like proper English

andlar (changes requested):
I think there's only one comment in here that's actually required to be addressed, and that's the one about the now-obsolete method under the @Then(CHPL API page title should be...) step regex. The rest of this is mostly formatting / style, and I'd appreciate it if it's changed, but not required.

andlar (approved):

ashwinimore (approved):

tmy1313 (approved):

PR #80, opened 2018-12-03
Add a missing controller and modify API endpoints to fix the failing …

iyerashwin14:
…tests


src/test/resources/Features/OCD-2262-RemoveDeprecatedApiPostCallsOfPutOrDelete.feature
```diff
@@ -13,18 +13,19 @@ Scenario Outline:
     Then list of "<API Endpoints>" operations should be displayed without deprecated calls 
     Examples: 
         | Controller                    |API Endpoints|
+        | chpl-file-controller          |Retrieve an API Documenation file--Upload an API Documenation file--Retrieve details about an API Documenation file|
```
andlar: Documentation isn't spelled correctly. My guess is that the API itself has the typo in it, but if that's the case then I think this code should have the correct spelling, and it should be up to a developer to fix the API docs to meet the tests. I wouldn't generally care too much about mis-spellings, but this is in our public documentation, and external users use this; I don't want us to look like we're not paying attention to the details

ashwinimore (approved):

andlar (changes requested):

kekey1 (approved):

andlar (approved):

PR #81, opened 2018-12-05
fix failing CMS ID test

ashwinimore:


src/test/java/gov/healthit/chpl/aqa/pageObjects/CMSidReverseLookupPage.java
```diff
@@ -56,17 +52,16 @@ public static WebElement generateCmsIdButton(final WebDriver driver) {
      * @return element holding main content
      */
     public static WebElement mainContent(final WebDriver driver) {
-        element = driver.findElement(By.id("main-content"));
-        return element;
+        return driver.findElement(By.id("main-content"));
```
andlar: I wonder if this `mainContent` element could move to the `BasePage` class. It _should_ be on all pages, or at least, I think it should, based on what I think we've done wrt 508 stuff...

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
```diff
@@ -2,45 +2,36 @@
...
 import cucumber.api.java.en.When;
-
 import gov.healthit.chpl.aqa.pageObjects.CMSidReverseLookupPage;
 import gov.healthit.chpl.aqa.pageObjects.ListingDetailsPage;
 
 /**
  * Class CMSIDsLookupSteps definition.
  */
-public class CMSIDsLookupSteps {
+public class CMSIDsLookupSteps extends Base {
```
andlar, line 17: Yay! OOP!

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
```diff
@@ -61,30 +52,31 @@ public void lookUpCmsId(final String cmsId) {
...
      * Generate a CMS ID.
+     * @throws Exception
      */
     @When("^I generate a CMS ID$")
     public void generateCmsId() {
-        WebDriverWait wait = new WebDriverWait(driver, TIMEOUT);
-        CMSidReverseLookupPage.generateCmsIdButton(driver).click();
-        wait.until(ExpectedConditions.visibilityOf(CMSidReverseLookupPage.cmsIdResults(driver)));
-    }
+       ((JavascriptExecutor) getDriver()).executeScript("arguments[0].removeAttribute('disabled','disabled')", CMSidReverseLookupPage.generateCmsIdButton(getDriver()));
```
andlar: Whoa!

This looks like a step that's making an element clickable when the page doesn't think it should be. That seems...improper

iyerashwin14 (approved):

src/test/resources/Features/OCD-2062-CMSIDsLookup.feature
```diff
@@ -1,6 +1,6 @@
 @Regression
 Feature: OCD-2062: Verify that newly generated & incorrectly removed CMS IDs still work
-  Some CMS IDs were lost during an outage. They need to be re-inserted, and beable to be validated against.
+  Some CMS IDs were lost during an outage. They need to be re-inserted, and made available for validation.
```
andlar, line 3: I did a `Ctrl-f` on "made av" and it found nothing. Does that step already exist? Or is this change missing that step?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
ashwinimore: I can see that being a bad practice in general but in this specific case, I know the element should be 'enabled'. I tested with isEnabled() method which returned false. 'Get cert ID' button was displayed but not enabled and it won't receive a click and the assert was failing as ID won't generate. If it can be enabled with some update, I'll change the test back to simple click event.

src/test/resources/Features/OCD-2062-CMSIDsLookup.feature
ashwinimore: This was just text edit in feature description - additional info. about feature. Scenario steps were not altered.

andlar: Good grief. I totally missed that this wasn't a step. That's what happens when you review things at 11:30 at night...

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CMSIDsLookupSteps.java
andlar: Agree that it should be clickable, but my guess is that the system is checking to see that the supplied Listings are enough to generate a CMS ID. The test should have a Wait condition until the item is clickable, to allow the system to process the addition of that listing and let the user know that it makes up a complete CMS-valid set of listings

andlar: If something changed in the validation logic for a CMS id and this set of listings became not-valid for a CMS id (or one of the listings was edited in such a way to make it invalid) we'd want this test to not be able to click, because it wouldn't be valid

ashwinimore: I did try wait before click event, it didn't work with even the explicit wait. The issue was/is with visibility condition and not load time. Let me look into alternate ways to locate and access this element.

andlar: If it's a visibility issue, not a clickable issue, then you should be able to use the javascript.execute(something-something[0].click()) thing

ashwinimore: Tried that, didn't help

andlar (approved):

PR #88, opened 2018-12-19
OCD 2579: Add creation date to various download files

iyerashwin14:


node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
andlar: All of the files in the `node_modules/` directory should be ignored. You'll need to, locally, `git rm node_modules/.yarn-integrity`, commit it (I think the best type might be "build") and continue to get rid of this file. If it's not auto-generated you should copy it somewhere else first, then after removing it from git you can return it here, if it's necessary, though it's probably auto-generated. My version of this file is very different

src/test/java/gov/healthit/chpl/aqa/pageObjects/CollectionsPages.java
```diff
@@ -41,4 +41,12 @@ public static WebElement cmsFaqLink(final WebDriver driver) {
         return element;
     }
 
+    /**
+     * Returns element holding SED Details File.
+     * @param driver WebDriver
```
andlar: In the other CHPL projects we're moving away from requiring javadocs on super simple methods like this one. If you want to do the same thing you can add `<property name="minLineCount" value="10" />` to the Checker -> TreeWalker -> JavadocMethod element of the Checkstyle file

src/test/java/gov/healthit/chpl/aqa/pageObjects/CollectionsPages.java
```diff
@@ -41,4 +41,12 @@ public static WebElement cmsFaqLink(final WebDriver driver) {
...
+     * @return element holding SED Details File
+     */
+    public static WebElement sedDetailsFileButton(final WebDriver driver) {
+        return driver.findElement(By.xpath("//a[@type='button']"));
```
andlar: This seems like a very broad xpath...

src/test/resources/Features/OCD-2023-monitorDownloadFiles.feature
```diff
@@ -30,10 +30,17 @@ Feature: OCD-2023: Monitor that download files on CHPL download resources page a
...
-  Scenario: At any given time, when downloaded, Surveillance Activity file is up-to-date
-    When I download the Surveillance Activity file
-    Then the downloaded file shows surveillance-all.csv filename
-
-  Scenario: At any given time, when downloaded, Non-Conformities file is up-to-date
-    When I download the Non-Conformities file
+ Scenario: At any given time, when downloaded, Non-Conformities file is up-to-date
+    When I select the Non-Conformities file from drop-down
+    And I download "surveillance-with-nonconformities" file
     Then the downloaded file shows surveillance-with-nonconformities.csv filename
```
andlar: This file should also have the  yyyymmdd extension. At the end of this ticket, _all_ of the files in that download dropdown should

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CollectionsPagesSteps.java
```diff
@@ -63,4 +63,25 @@ public void verifyCMSFAQLink(final String cmsfaqLink) {
         assertTrue(link.contains(cmsfaqLink), "Expect " + cmsfaqLink + " to be found in " + link);
     }
 
+    /**
+     * Get user to the Collections page.
+     */
+    @Given("^I am on the Collections page$")
+    public void iAmOnCollectionsPage() {
```
andlar: This step is a less useful one than the one in `loadCollectionsPage`, and unclear. What "collections" page? There are five or six of them, and this one loads a single one without being clear that it's doing that

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -320,6 +313,33 @@ public void checkifDownloadDirectoryisEmpty() throws IOException {
...
+    public void surveillanceActivityFileEndsWithDateTimeFormat(final String fileName, final String dateTimeFormat) {
+        File[] filenames = Hooks.getDownloadDirectory().listFiles();
+        for (File file : filenames) {
+            String dwldFileName = file.getName();
+            if (dwldFileName.startsWith(fileName)) {
+                String[] fileTokens = dwldFileName.split("-");
+                String extToken = fileTokens[fileTokens.length - 1];
+                String[] dTokens = extToken.split("\\.");
+                String fileDate = "-" + dTokens[0];
+                SimpleDateFormat df = new SimpleDateFormat(dateTimeFormat);
```
andlar, line 307: The dateTimeFormat value that's passed in is `-yyyMMdd_HHmmss`, including that initial `-`. Does that get correctly converted to a SimpleDateFormat? Even if it does work, it shouldn't

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CollectionsPagesSteps.java
```diff
@@ -63,4 +63,25 @@ public void verifyCMSFAQLink(final String cmsfaqLink) {
...
+        WebDriverWait wait = new WebDriverWait(driver, TIMEOUT);
+        wait.until(ExpectedConditions.visibilityOf(CollectionsPages.mainContent(driver)));
+    }
+
+    /**
+     * Download SED Details.csv file.
+     * @param fileName expected file name
+     */
+    @When("^I download the \"([^\"]*)\" file$")
+    public void downloadSurveillanceActivityFile(final String fileName) {
```
andlar: Function name suggests surveillance. First line of the method looks like SED. Step definition doesn't care. I'm not sure what this is supposed to do

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CollectionsPagesSteps.java
```diff
@@ -63,4 +63,25 @@ public void verifyCMSFAQLink(final String cmsfaqLink) {
...
+        WebElement link = CollectionsPages.sedDetailsFileButton(getDriver());
+        ((JavascriptExecutor) getDriver()).executeScript("arguments[0].click();", link);
+        super.downloadFile(fileName);
```
andlar: Does this method download the file? Or does the super method download the file?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -320,6 +313,33 @@ public void checkifDownloadDirectoryisEmpty() throws IOException {
         assertFalse(Hooks.getDownloadDirectory().list().length > 0, "directory is not empty");
     }
 
+    /**
+     * Assert filename ends with date and time stamp.
+     * @param dateTimeFormat expected Date and Time
+     * @param fileName expected File name
+     */
+    @Then("^the \"([^\"]*)\" file ends with \"([^\"]*)\"$")
+    public void surveillanceActivityFileEndsWithDateTimeFormat(final String fileName, final String dateTimeFormat) {
```
andlar: Why does this method say "surveillance" in the name, when it looks like it's supposed to be generic?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -61,4 +61,28 @@ public WebDriverWait getWait() {
...
     }
+
+    /**
+     * Check whether the file is downloaded.
+     * @param fileName expected File name
+     */
+    public void downloadFile(final String fileName) {
+        String downloadFileName = null;
+        boolean fileFound = false;
+        final long sleepTime = 5 * 1000;
```
andlar: Checkstyle should have complained about magic numbers here

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -61,4 +61,28 @@ public WebDriverWait getWait() {
     public void setWait(final WebDriverWait wait) {
         this.wait = wait;
     }
+
+    /**
+     * Check whether the file is downloaded.
+     * @param fileName expected File name
+     */
+    public void downloadFile(final String fileName) {
```
andlar: The name of the method does not match what the method does, nor what the documentation says

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -61,4 +61,28 @@ public WebDriverWait getWait() {
     public void setWait(final WebDriverWait wait) {
         this.wait = wait;
     }
+
+    /**
+     * Check whether the file is downloaded.
```
andlar: This is not a _complicated_ method, but the documentation is incomplete. It not only checks whether the file is downloaded, it waits (forever) until that file exists. As well as updating the documentation, you should probably also put some sort of "it's been three hours of waiting, it's never going to happen, throw an exception" check in here (probably with less than three hours)

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -258,36 +258,29 @@ public void theDownloadFileHasAtLeastItems(final String itemParam) {
     }
 
     /**
-     * Select Surveillance Activity file from drop down and download .csv file.
+     * Select Surveillance Activity file from drop-down.
      */
-    @When("^I download the Surveillance Activity file$")
-    public void downloadSurveillanceActivityFile() {
+    @When("^I select the Surveillance Activity file from drop-down$")
+    public void selectSurveillanceActivityFile() {
```
andlar: All of these steps should be combined into one, with a parameter for the name of the file as it's shown in the download drop-down. The Page Object elements should be consolidated into a single one as well, where instead of going to the "nth" item in the select, it gets one based on the visible name

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/pageObjects/CollectionsPages.java
ashwinimore: I'd like to request id for SED Download button on SED Info collections page.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CollectionsPagesSteps.java
ashwinimore: There is an existing step in this file (see line 50) that can be used in new scenario as - 'Given I am on "SED Information for 2015 Edition" collections page: "sed"', no need for new step and step definition.

ashwinimore (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
iyerashwin14: I'm reusing the existing code logic as such editor does not give any Checkstyle error

ashwinimore (approved):

chpl-checkstyle-checks.xml
```diff
@@ -68,6 +68,7 @@
         <module name="JavadocMethod">
             <property name="severity" value="warning"/>
             <property name="scope" value="public" />
+			<property name="minLineCount" value="10" />
```
andlar: This looks like a file with mixed tabs and spaces

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -258,51 +260,17 @@ public void theDownloadFileHasAtLeastItems(final String itemParam) {
...
+     * @param fileName expected name of the file in drop-down list
+     * @param downloadedFileName expected downloaded file name
+     * @throws FileNotFoundException if the expected file not found
      */
-    @When("^I download the Surveillance Activity file$")
-    public void downloadSurveillanceActivityFile() {
-        ChplDownloadPage.downloadoptionSurveillanceFile(getDriver()).click();
+    @When("^I select the \"([^\"]*)\", download it from drop-down and I see the \"([^\"]*)\"$")
+    public void selectAndDownloadFileFromDropdown(final String fileName, final String downloadedFileName) throws FileNotFoundException {
+        WebElement link =   ChplDownloadPage.selectFilefromDropdown(getDriver(), fileName);
```
andlar: Why the extra spaces after the `=`?

src/test/resources/Features/OCD-2023-monitorDownloadFiles.feature
```diff
@@ -1,39 +1,47 @@
...
+    Then the "<Downloaded File Name>" ends with "yyyyMMdd_HHmmss" 
+    Examples: 
+        |File                 |Downloaded File Name             |
+        |Surveillance Activity|surveillance-all                 |
+        |Non-Conformities     |surveillance-with-nonconformities|
 
-  Scenario: At any given time, when downloaded, Non-Conformities file is up-to-date
-    When I download the Non-Conformities file
-    Then the downloaded file shows surveillance-with-nonconformities.csv filename
+Scenario:
```
andlar: Why is this scenario not one of the outlined ones above? It's the same logic, right?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CollectionsPagesSteps.java
```diff
@@ -17,7 +19,7 @@
 /**
  * Class CollectionsPagesSteps definition.
  */
-public class CollectionsPagesSteps {
+public class CollectionsPagesSteps extends Base {
     private WebDriver driver;
```
andlar: If this file extends Base, then it shouldn't need any of these private variables, right?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
andlar: You might be re-using logic, but according to git this is all new code. In any case, whether or not checkstyle complains about it, it's a case where you have magic numbers, and we don't like magic numbers, so it should be changed

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -61,4 +63,35 @@ public WebDriverWait getWait() {
...
+        int retryCount = 0;
+        while (!fileFound && retryCount <= MAX_RETRYCOUNT) {
+            File[] files = Hooks.getDownloadDirectory().listFiles();
+            for (File file : files) {
+                downloadFileName = file.getName();
+                if (downloadFileName.startsWith(fileName)) {
+                    fileFound = true;
+                    break;
+                }
+                try {
```
andlar: I think this try-catch block sleeps between checking each file. Shouldn't it be moved out of the for loop to be part of the while?

andlar (changes requested):

src/test/resources/Features/OCD-2023-monitorDownloadFiles.feature
iyerashwin14: I have this as a separate scenario because the chpl-sed file is on the Collections page whereas the other files are on the Resources page. 
Yes, it's the same logic!

src/test/resources/Features/OCD-2023-monitorDownloadFiles.feature
```diff
@@ -1,39 +1,47 @@
...
+        | PROD |    2014 | csv  |    1 | 2900 | KB    |  5190 |
+        | PROD |    2015 | xml  |    1 |  150 | MB    |   590 |
+        | PROD |    2015 | csv  |    1 |  350 | KB    |   590 |
 
-  Scenario: At any given time, when downloaded, Surveillance Activity file is up-to-date
-    When I download the Surveillance Activity file
-    Then the downloaded file shows surveillance-all.csv filename
+Scenario Outline: 
+    At any given time, when downloaded, file name ends with Date and time format 
+    When I select the "<File>", download it from drop-down and I see the "<Downloaded File Name>"
```
andlar, line 37: Doesn't this need a `Given` for the page and environment? How does the scenario know what page it should be on?

src/test/resources/Features/OCD-2023-monitorDownloadFiles.feature
```diff
@@ -1,39 +1,47 @@
...
+        |File                 |Downloaded File Name             |
+        |Surveillance Activity|surveillance-all                 |
+        |Non-Conformities     |surveillance-with-nonconformities|
 
-  Scenario: At any given time, when downloaded, Non-Conformities file is up-to-date
-    When I download the Non-Conformities file
-    Then the downloaded file shows surveillance-with-nonconformities.csv filename
+Scenario: 
+    At any given time, when downloaded, SED Details file name ends with Date and time format 
+    Given I am on "SED Information for 2015 Edition" collections page: "sed"
```
andlar: The other download file checks have outlines that let the test check on each environment. I think this one should too

src/test/resources/Features/OCD-2023-monitorDownloadFiles.feature
```diff
@@ -1,39 +1,47 @@
-@Regression
-Feature: OCD-2023: Monitor that download files on CHPL download resources page are correct and current
-  Verify download files are recent, to ensure file generation didn't fail at any point.
+@Regression
```
andlar, line 1: I don't know if checkstyle knows about .feature files, but github shows some trailing spaces added to these files. Request looking through these and cleaning those up

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
```diff
@@ -320,6 +288,35 @@ public void checkifDownloadDirectoryisEmpty() throws IOException {
...
+            if (dwldFileName.startsWith(fileName)) {
+                String[] fileTokens = dwldFileName.split("-");
+                String extToken = fileTokens[fileTokens.length - 1];
+                String[] dTokens = extToken.split("\\.");
+                String fileDate = dTokens[0];
+                SimpleDateFormat df = new SimpleDateFormat(dateTimeFormat);
+                df.setLenient(false);
+                try {
+                    df.parse(fileDate);
+                    assertTrue(true);
```
andlar: Is there a `pass` method similar to the `fail` ones from a few lines down? This "assertTrue(true)" thing gives me a headache. If there is such a pass method it'd be much nicer to see something like `pass("Filename " + dwldFileName + " has correct date time format");` or something like it

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -61,4 +64,35 @@ public WebDriverWait getWait() {
...
+
+    /**
+     * Checks whether the specified file is downloaded or not by iterating 8 times where each iteration has a wait of 5 seconds.
+     * @param fileName expected downloaded file name
+     * @throws FileNotFoundException if the expected file not found
+     */
+    public void checkIfFileIsDownloaded(final String fileName)throws FileNotFoundException {
+        String downloadFileName = null;
+        boolean fileFound = false;
+        final long sleepTime = SLEEP_TIME;
```
andlar: `SLEEP_TIME` is already defined above; it doesn't need to be redefined here

chpl-checkstyle-checks.xml
```diff
@@ -68,6 +68,7 @@
         <module name="JavadocMethod">
             <property name="severity" value="warning"/>
             <property name="scope" value="public" />
+		 <property name="minLineCount" value="10" />
```
andlar: Indentation is still messed up here

andlar (changes requested):

src/test/resources/Features/OCD-2023-monitorDownloadFiles.feature
iyerashwin14: I'm reusing an existing step suggested in the below code review comment by Ashwini : -
"There is an existing step in this file (see line 50) that can be used in new scenario as - 'Given I am on "SED Information for 2015 Edition" collections page: "sed"', no need for new step and step definition."
I have two options: - change the existing step definition (might affect the existing features) or add a new one! So what would you suggest?

andlar: One of the reasons for these tests is to find out when the various environments don't have up to date files. The best way to do that is to parameterize the environments and test against them all. You could expand on the existing step to add a param for the environment, or write a new step. Either would be fine, though the first one would ripple through to some other number of tests

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplDownloadSteps.java
iyerashwin14: I searched if there is any such API available to assert but couldn't find any.

andlar: I just realized we're missing the whole point. The reason for putting the datestamp in the file name is to assert that it's not too old. Look at the `theDownloadFileIsNotOld` method to see how that one parses the filename's datestamp and checks against it being too old. Then there can be an actual `assertTrue(something something something)` that makes sense, and actually verifies what we're looking for in the first place

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CollectionsPagesSteps.java
```diff
@@ -48,9 +34,21 @@ public void openAcbFilterOptions() {
...
+        wait.until(ExpectedConditions.visibilityOf(CollectionsPages.mainContent(getDriver())));
+    }
+
+    /**
+     * Loads collections pages.
+     * @param pname to get page name in url
+     * @param tEnv test environment in which tests will be run
+     */
+    @Given("^I am on collections page: \"([^\"]*)\" on \"([^\"]*)\"$")
+    public void loadCollectionPage(final String pname, final String tEnv) {
```
andlar: I don't like having two methods with almost exactly the same name and the same signature, but I don't know what the right solution is

chpl-checkstyle-checks.xml
```diff
@@ -68,6 +68,7 @@
         <module name="JavadocMethod">
             <property name="severity" value="warning"/>
             <property name="scope" value="public" />
+	    <property name="minLineCount" value="10" />
```
andlar: This is still misaligned

iyerashwin14: I tried editing with different editors but at my end, the alignment looks good. Can you please help me out with this?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CollectionsPagesSteps.java
```diff
@@ -48,9 +34,21 @@ public void openAcbFilterOptions() {
...
+        wait.until(ExpectedConditions.visibilityOf(CollectionsPages.mainContent(getDriver())));
+    }
+
+    /**
+     * Loads collections page in test environment.
+     * @param pageName to get page name in url
+     * @param tEnv test environment in which tests will be run
+     */
+    @Given("^I am on collections page: \"([^\"]*)\" on \"([^\"]*)\"$")
+    public void loadCollectionsPageInTestEnv(final String pageName, final String tEnv) {
```
andlar: This says load page on test environment, but it's called to load the page in three different environments, including PROD. Technically, we don't really even _have_ a test environment

iyerashwin14: Correct, that's true! Can you please suggest some name?

iyerashwin14: Can you please suggest from the following names:-
browseCollectionsPageInEnv
openCollectionsPageInEnv

src/test/resources/Features/OCD-2114-SLIFilterCheckedByDefault.feature
```diff
@@ -6,14 +6,14 @@ Scenario: Verify SLI Compliance filter option under ACB filters is checked by de
 	Then I see that "SLI Compliance" checkbox is checked
 
 Scenario Outline: Verify SLI Compliance filter option is checked by default on collections pages
-	Given I am on "<PageTitle>" collections page: "<PageName>"
+	Given I am on collections page: "<PageName>" on "DEV"
```
andlar, line 9: I don't like that these steps are now specific to DEV. If I wanted to run this test locally I'd have to change the feature file, and that's not okay. That said, I'm not going to hold up the approval any longer on this because even if it's not 100% correct it's close enough to make progress on

andlar (approved):

PR #93, opened 2019-01-15
feat: develop download functionality for search results

iyerashwin14:


src/test/resources/Features/OCD-111-developDownloadFunctionalityForSearchResults.feature
```diff
@@ -0,0 +1,24 @@
...
+    Examples:
+      |CHPL ID   |
+      |CHP-022844|
+      |CHP-026059|
+      |CHP-028979|
+
+Scenario Outline: Verify whether correct error message is displayed if more than 50 results per page is selected 
+    Given I am on CHPL search page
+    When I search "<count>" results per page and I try to download the file
+    Then I see the text as "Please reduce results to less than 50 to download them"
```
andlar: I don't like the name of this step. It's so very generic that if you saw a list of steps you'd have no idea what area of the site this was referring to

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -95,6 +95,38 @@ public void checkIfFileIsDownloaded(final String fileName)throws FileNotFoundExc
...
 
+    /**
+     * Checks if the file contains all the data and is downloaded completely or not.
+     * @param fileName expected downloaded file name
+     * @throws FileNotFoundException if the expected file not found
+     */
+    public void checkCompleteFileDownload(final String fileName) throws FileNotFoundException {
+        boolean completeFileDwnld = true;
+        int retryCount = 0;
+        Long size = new Long(-1);
```
andlar: What is this used for? It gets set, and checked against, but is never used

src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
```diff
@@ -164,4 +164,24 @@ public static WebElement selectAllStatus(final WebDriver driver) {
...
+        return element;
+    }
+
+    public static WebElement searchResultPerPage(final WebDriver driver) {
+        element = driver.findElement(By.xpath("//select[@id='pageSizeTop']"));
+        return element;
+    }
+
+    public static WebElement searchResultText(final WebDriver driver) {
+        element = driver.findElement(By.xpath("//span[@class='ng-binding ng-scope']"));
```
andlar: Does this actually work? I feel like this could return dozens of elements

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
ashwinimore: I agree. That class name isn't unique. I know at least one other element that uses same class name ( i think announcements text). It worked in this case because of it's order on page. It's at top, so was found first and worked.
I'd recommend using alternative relative xpath.

src/test/resources/Features/OCD-111-developDownloadFunctionalityForSearchResults.feature
ashwinimore: Proposed:
Given I am on search page
And I scroll down to select <count> results per page for results display
When I click Download Results button in search filters to download the search results
Then I see that download for >50 count is not allowed and alert <text> is displayed

ashwinimore (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
iyerashwin14: It's been used to make sure whether the complete file is downloaded

andlar: Where? What line?

iyerashwin14: Refer line number : 111
Here is the code snippet:
 Long downloadFileSize = file.length();
                    if (size < downloadFileSize) {
                        size = downloadFileSize;
                    } else {
                        completeFileDwnld = false;
                        break;
                    }

andlar: I'm trying to follow this logic. In pseudocode, what you're doing is:

```
while I haven't completed the download, and haven't reached the max retry value:
  for each file in the download directory:
    get the size of the file
    if the size of the last file is >= the new file's size
      stop looking at the files, and say we've found our file
    otherwise
      wait for 5 seconds and check the next file
If I havnen't said we found the file
  report that it wasn't found
```

* Why pass in the name of the file if you're not checking it against the file that was found?
* What would happen if there was more than one file in this directory?
* Why is `completeFileDwnld` true at the start of the method, and false at the end, assuming everything worked correctly? It seems to me that "Complete File Downloaded" should be true at the end, not the beginning

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -95,6 +95,38 @@ public void checkIfFileIsDownloaded(final String fileName)throws FileNotFoundExc
         }
     }
 
+    /**
+     * Checks if the file contains all the data and is downloaded completely or not.
+     * @param fileName expected downloaded file name
+     * @throws FileNotFoundException if the expected file not found
+     */
+    public void checkCompleteFileDownload(final String fileName) throws FileNotFoundException {
+        boolean completeFileDwnld = true;
```
kekey1: Could you just write out the word "download" in this variable name? I don't feel like leaving a few letters out is advantageous for any particular reason.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
kekey1: Trying to follow this as well.. maybe if you changed the while condition to "!completeFileDownloaded and retryCount <= MAX_RETRY_COUNT" then the booleans could be switched to make more sense as Andrew described.

kekey1 (changes requested):

kekey1 (approved):

ashwinimore (approved):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
andlar: In the download file folder, what would happen if there were two files there? Let's say there's already a file called "chpl-2015.xml" with a size of 100Mb, and we start downloading the sed-details file. Assuming it doesn't download immediately, it's name would be something like "sed-details.crdownload", and let's assume it's going to be 30Kb, but gets 5Kb every 5 seconds. Here's what it looks like on each run through the for loop

Assuming chpl-2015 is first
```
+-----+----------------+--------+---------------------------------+-------------------------------------------+
|size |downloadFileSize|download|file name                        |State                                      |
+-----+----------------+--------+---------------------------------+-------------------------------------------+
|-1   |undefined       |false   |undefined                        |initial state                              |
+-----+----------------+--------+---------------------------------+-------------------------------------------+
|100Mb|100Mb           |false   |chpl-2015.xml                    |first file check, now sleep for 5 sec      |
+-----+----------------+--------+---------------------------------+-------------------------------------------+
|100Mb|10Kb            |true    |sed-details.crdownload           |"Found" file, but it's the wrong one       |
+-----+----------------+--------+---------------------------------+-------------------------------------------+
```

Assuming sed-details is first

```
+-----+----------------+--------+---------------------------------+--------------------------------------------+
|size |downloadFileSize|download|file name                        |State                                       |
+-----+----------------+--------+---------------------------------+--------------------------------------------+
|-1   |undefined       |false   |undefined                        |initial state                               |
+-----+----------------+--------+---------------------------------+--------------------------------------------+
|5Kb  |5Kb             |false   |sed-details.crdownload           |first file check, now sleep for 5 sec       |
+-----+----------------+--------+---------------------------------+--------------------------------------------+
|100Mb|100Mb           |false   |chpl-2015.xml                    |Second file check, now sleep for 5 more sec |
+-----+----------------+--------+---------------------------------+--------------------------------------------+
|100Mb|15Kb            |true    |sed-details.crdownload           |Found it again, but also the wrong one      |
+-----+----------------+--------+---------------------------------+--------------------------------------------+
```

iyerashwin14: @andlar There won't be more than one file in the directory because the AND condition (clears the directory) is executed before this step so there is just one file in the download directory

iyerashwin14: The actual logic that I have used is like:- 
while I haven't completed the download, and haven't reached the max retry value:
  for each file in the download directory: 
    get the size of the file
  if the size of the current file is > size variable
	Sleep to allow the file to be downloaded completely that doesn't allow the browser to close
    Otherwise
          Size variable = current file size
    Then the file is completely downloaded
     Else If
	After going through all the retry counts and still, the file is not downloaded 
     Throw exception
	File not found

iyerashwin14: AND code snippet that I'm reusing:-
    /**
     * Print filenames if directory is not empty.
     * Assert that download directory is empty
     * @throws IOException if unable to clean the directory
     */
    @And("^the download directory is empty$")
    public void checkifDownloadDirectoryisEmpty() throws IOException {
        System.out.println("Clearing directory: " + Hooks.getDownloadDirectory().getAbsolutePath());
        for (String fileName : Hooks.getDownloadDirectory().list()) {
            System.out.println("Removing: " + fileName);
        }
        FileUtils.cleanDirectory(Hooks.getDownloadDirectory());
        assertFalse(Hooks.getDownloadDirectory().list().length > 0, "directory is not empty");
    }

src/test/resources/Features/OCD-111-developDownloadFunctionalityForSearchResults.feature
```diff
@@ -5,7 +5,7 @@ Feature: OCD-111 Develop export/download functionality for search results
 Scenario Outline: Verify whether search results file is downloaded and displays correct information 
     Given I am on CHPL search page
     And the download directory is empty 
-    When I click Download Search Results button
+    When I wait for "<CHPL ID>" to load in the webpage and I click Download Search Results button
```
andlar: This step definition shouldn't change

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ChplSearchSteps.java
```diff
@@ -172,12 +171,15 @@ public void selectAllCertStatusFilters() {
...
      */
-    @When("^I click Download Search Results button$")
-    public void clickOnDownloadSearchResultsButton() {
+    @When("^I wait for \"([^\"]*)\" to load in the webpage and I click Download Search Results button$")
+    public void clickOnDownloadSearchResultsButton(final String searchResultsChplID) {
         SearchPage.browseButton(getDriver()).click();
         WebElement link = SearchPage.downloadsearchResultsButton(getDriver());
         ((JavascriptExecutor) getDriver()).executeScript("arguments[0].click();", link);
+        WebDriverWait wait = new WebDriverWait(getDriver(), TIMEOUT);
+        wait.until(ExpectedConditions.visibilityOf(SearchPage.loadChplID(getDriver(), searchResultsChplID)));
```
andlar, line 180: I like the wait step, but it should be _before_ you click the button to download the data. A user will likely wait until the search has finished, and that's the behavior we're trying to mimic with these tests

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -108,35 +107,31 @@ public void checkCompleteFileDownload(final String fileName, final String extens
...
-            for (File file : fileFilterList) {
-                Long tempModifiedTime = file.lastModified();
-                if (lastModifiedTime <= tempModifiedTime) {
-                    lastModifiedTime = tempModifiedTime;
-                    recentModifiedFile = file;
+                download = false;
+                retryCount = 0;
+                while (!download && retryCount <= MAX_RETRYCOUNT) {
+                    Long downloadFileSize = files[i].length();
+                    if (size < downloadFileSize) {
```
andlar: Why are you still using the file size for anything? I thought we were getting rid of that?

The logic should be something like this (in pseudocode):
```
while (!foundFIle && retryCount <= MAX_RETRY_COUNT)
    get files
        for each file
            if (file matches name restrictions
                && (Now() - file.lastModifiedDate) < (SLEEP_TIME * 3)) {
                foundFile = true;
                break; //optional, only really needed if we're expecting to find hundreds or thousands of files in this directory
                    }
    sleep(SLEEP_TIME)
    retryCount += 1
if (!foundFile)
    throw new FileNotFoundException("File not downloaded");
```

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -17,7 +17,8 @@
     protected static final long TIMEOUT = 30;
     protected static final long LONG_TIMEOUT = 120;
     private static final int MAX_RETRYCOUNT = 8;
-    private static final int SLEEP_TIME = 5000;
+    private static final int COUNT = 3;
```
andlar: What's this counting?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -95,6 +96,37 @@ public void checkIfFileIsDownloaded(final String fileName)throws FileNotFoundExc
...
+                File[] files = Hooks.getDownloadDirectory().listFiles();
+                for (int i = 0; i < files.length; i++) {
+                    Long currentTime = System.currentTimeMillis();
+                    if ((files[i].getName().startsWith(fileName) && files[i].getName().endsWith(extension)) && ((currentTime - files[i].lastModified()) < SLEEP_TIME * COUNT)) {
+                        foundFile = true;
+                        break;
+                    }
+                }
+                retryCount++;
+                Thread.sleep(SLEEP_TIME);
```
andlar: Probably should surround this with an "if !foundFile" so if you do find it we don't needlessly wait another 5 seconds

andlar (changes requested):
Almost there!

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
iyerashwin14: (Now() - file.lastModifiedDate) < (SLEEP_TIME * 3)) 
As per your suggestion, I have implemented this logic just considering the count as 3 which means the file should not be older than 15 seconds

andlar: Yes, I get what it's actually doing. What I'm trying to point out is that the variable name `COUNT` is not descriptive at all. It's not actually counting anything, and even if it were, it'd be much more useful to know what it was counting

andlar (approved):

PR #95, opened 2019-01-23
test: verify API documentation updates to endpoint details

iyerashwin14:


ashwinimore (approved):

src/test/java/gov/healthit/chpl/aqa/asserts/ChplAPIPageAsserts.java
```diff
@@ -64,4 +73,46 @@ public void verifyContentUnderCertifiedHealthITProductListingAndURL(final String
...
+
+    /**
+     * Assert that implementation notes for the endpoint link is updated.
+     * @throws Exception if File not found
+     */
+    @Then("^API endpoints should show updated implementation notes$")
+    public void apiEndpointsImplementationNotesDisplaysUpdatedText() throws Exception {
+        JSONParser jsonParser = new JSONParser();
+        Boolean failedCase = false;
+        Path resourceDirectory = Paths.get("src", "test", "resources", "implementationNotes.json");
```
kekey1: I usually try not to hardcode paths of files... src/test/resources is typically on the classpath by default so Java can find files in there without you having to tell it the path. Can you try reading this file in another way? Something like 
` File file = new File(getClass().getClassLoader().getResource("implementationNotes.json").getFile())`
Also, could you make the filename "implementationNotes.json" a constant instead of hardcoded?

kekey1 (changes requested):

kekey1 (approved):

PR #99, opened 2019-01-31
test: update wait condition to use common locator found in 2014 and 2015 ed. listings

ashwinimore:


src/test/java/gov/healthit/chpl/aqa/pageObjects/ListingDetailsPage.java
```diff
@@ -327,6 +327,15 @@ public static WebElement surveillanceActivitiesAccordion(final WebDriver driver)
         return driver.findElement(By.id("details-surveillance-activities"));
     }
 
+    /**
+     * Returns Targeted User data from additional Info panel.
+     * @param driver WebDriver
+     * @return the text element
+     */
+    public static WebElement targetedUser(final WebDriver driver) {
+        return driver.findElement(By.xpath("//*[@id=\"panel-additional-information\"]/span[2]/ul/li"));
```
andlar: I updated the website to have some IDs for these areas. They're in this PR: https://github.com/chpladmin/chpl-website/pull/405

The reason is that the position of all of these items is dependent on which other items exist in the Listing, so you can't just count the 2nd or 3rd ones. Recommend changing both this and the meaningfulUse page object definition

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ListingDetailsSteps.java
```diff
@@ -75,7 +75,7 @@ public void iLookAtSEDDetails() {
     public void iLookAtAdditionalInformation() {
         WebElement link = ListingDetailsPage.additionalInfoAccordion(getDriver());
         ((JavascriptExecutor) getDriver()).executeScript("arguments[0].click();", link);
-        getWait().until(ExpectedConditions.visibilityOf(ListingDetailsPage.targetedUser(getDriver())));
+        getWait().until(ExpectedConditions.visibilityOf(ListingDetailsPage.meaningfulUseUsers(getDriver())));
```
andlar, line 78: This is the only item that is guaranteed to be on every listing's detail's additional information. Definitely recommend keeping this step as is :)

andlar (changes requested):

iyerashwin14 (approved):

src/test/java/gov/healthit/chpl/aqa/pageObjects/ListingDetailsPage.java
ashwinimore: Awesome. Thanks! I'll update locators once PR is merged.

andlar (approved):

PR #106, opened 2019-02-20
test: OCD-1566 - update default search parameters for recently retired ACBs

iyerashwin14:


src/test/resources/Features/OCD-1566-updateDefaultSearchParametersForRecentlyRetiredACBs.feature
```diff
@@ -0,0 +1,45 @@
...
+    And I mark it as retired on "08/01/2015"
+    Then ONC ACB have retired status as "Yes"
+    Examples:
+      |ACB Name      |
+      |InfoGard      |
+      |ICSA Labs     |
+      |Drummond Group|
+      |SLI Compliance|
+
+  Scenario Outline: Verify ACBs can be unretired
```
andlar: Something is weird with these scenarios. You shouldn't be able to unretire ACBs that are currently not retired. If these scenarios assume the "retire an ACB" scenario has happened then that's not okay

src/test/resources/Features/OCD-1566-updateDefaultSearchParametersForRecentlyRetiredACBs.feature
```diff
@@ -0,0 +1,45 @@
...
+      |ICSA Labs     |
+      |Drummond Group|
+      |SLI Compliance|
+
+  Scenario: Verify ACBs display an error message if any future date of retirement is selected 
+    Given I'm logged in as "ROLE_ADMIN"
+    When I navigate to ONC ACB Management page
+    And I click on "InfoGard"
+    And I edit an ACB 
+    And I mark it as retired on "08/01/2020"
```
andlar: What happens on August 2nd of 2020? I know that's a long time from now, but we want our system to be future proof. Maybe set the year to a hundred years from now, or maybe set the date to "now + 1 day", but either way, a specific date will fail before we're ready for it

andlar (changes requested):

src/test/resources/Features/OCD-1566-updateDefaultSearchParametersForRecentlyRetiredACBs.feature
iyerashwin14: If you see the earlier scenario, I have retired the ACBs so in this scenario I'm just marking it as unretired. So that in the next nightly AQA execution, the test will again retire the ACBs

src/test/resources/Features/OCD-1566-updateDefaultSearchParametersForRecentlyRetiredACBs.feature
iyerashwin14: I have just selected a future date that throws an error "Retirement date is required and must be before "now". 
I will set the year to a hundred years from now.

src/test/resources/Features/OCD-1566-updateDefaultSearchParametersForRecentlyRetiredACBs.feature
andlar: One of the basic ideas of a test is that it should never depend on any other tests having happened or not. Tests should be able to be run in any order, skipping any or all of them, and all still work correctly. If one test depends on another test then that test is wrong

ashwinimore: For the 'retire ACB' scenario, i don't think there's a need to test it for multiple ACBs. I think, testing it for one ACB should suffice. Another alternative approach to unretire ACB & get ready for next day run could be to unretire ACB followed by retire action in step, instead of adding a separate scenario for it. I  know it's not ideal but could work well.

ashwinimore (approved):

andlar (approved):

PR #121, opened 2019-04-09
test: OCD-2797-verify surveillance upload-confirm workflow is functional

ashwinimore:
next to do: delete uploaded surveillance (manageSurveillance page cached)


src/test/java/gov/healthit/chpl/aqa/asserts/DpmanagementUploadPageAsserts.java
```diff
@@ -36,4 +55,16 @@ public void verifyLongTestTaskIdAndParticipantIdErrorsOnUpload(final String erro
...
+     * Load listing details to verify surveillance was uploaded successfully.
+     */
+    @Then("^I see that surveillance was uploaded successfully to CHPL and listing details show surveillance activity$")
+    public void verifySurveillanceConfirmWasSuccessful() {
+        WebElement accordion = ListingDetailsPage.surveillanceActivitiesAccordion(getDriver());
+        ((JavascriptExecutor) getDriver()).executeScript("arguments[0].click();", accordion);
+
+        String surveillanceText = ListingDetailsPage.surveillanceActivitiesPanel(getDriver()).getText();
+
+        assertTrue(surveillanceText.contains("Open Surveillance, Began Sep 20, 2017: 1 Open and 2 Closed Non-Conformities Were Found"));
```
andlar: How will we know that the surveillance found here isn't the one that was uploaded "the day before yesterday", or whatever?

src/test/java/gov/healthit/chpl/aqa/asserts/DpmanagementUploadPageAsserts.java
```diff
@@ -36,4 +55,16 @@ public void verifyLongTestTaskIdAndParticipantIdErrorsOnUpload(final String erro
         assertTrue(actualString.contains(errorMessage), "Expect \"" + errorMessage + "\" to be found in \"" + actualString + "\"");
     }
 
+    /**
+     * Load listing details to verify surveillance was uploaded successfully.
+     */
+    @Then("^I see that surveillance was uploaded successfully to CHPL and listing details show surveillance activity$")
+    public void verifySurveillanceConfirmWasSuccessful() {
+        WebElement accordion = ListingDetailsPage.surveillanceActivitiesAccordion(getDriver());
+        ((JavascriptExecutor) getDriver()).executeScript("arguments[0].click();", accordion);
```
andlar: I don't know where to mention this, but if you go to the regular details page, but put "/surveillance" at the end of it, it'll open with that accordion open already: https://chpl.ahrqdev.org/#/listing/9672/surveillance

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ManageDevelopersAndProductsSteps.java
```diff
@@ -368,4 +369,70 @@ public void uploadListingWithBadInputsInTestTaskParticipantFields(final String c
     public void inspectListingDetails(final String chplId) {
         DpManagementPage.inspectButtonForBadDataListing(getDriver(), chplId).click();
     }
+
+    /**
+     * Upload a surveillance.
+     * @throws URISyntaxException
+     * @param inputChplId is a CHPL ID of a listing for which surveillance is uploaded
+     */
+    @When("^I upload a surveillance activity for listing with CHPL ID \"([^\"]*)\"$")
```
andlar: Where is the CHPL ID used in this method?

src/test/resources/Features/OCD-2797-surveillanceUpload.feature
```diff
@@ -0,0 +1,13 @@
...
+
+  Scenario: Upload and confirm surveillance activity
+    Given I'm logged in as "ROLE_ACB"
+    And I am on Upload Products & Surveillance page
+    When I upload a surveillance activity for listing with CHPL ID "15.05.05.2760.ISCD.01.00.1.181101"
+    Then I see upload successful message for surveillance upload
+    When I go to Confirm Pending Surveillance Activities Page
+    And I inspect surveillance activity details for listing with CHPL ID "15.05.05.2760.ISCD.01.00.1.181101"
+    And I confirm surveillance activity for listing with CHPL ID "15.05.05.2760.ISCD.01.00.1.181101"
+    Then I see that surveillance was uploaded successfully to CHPL and listing details show surveillance activity
```
andlar: I think this step should have the CHPL ID in it. I don't see where the navigation to the listing details page is happening in these steps so I'd assume it would be here, because the previous step doesn't suggest any navigation.

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/asserts/DpmanagementUploadPageAsserts.java
ashwinimore: I had added a note when creating a PR, along with commit message but it's not showing. Next to do is delete uploaded surveillance from this listing so there's no previous surveillance activity with same dates and NC counts. manageSurveillance page is cached and surveillance activity doesn't load/show unless I open a new session in different browser, cache refresh in same session didn't help either. I need to figure out a way to access it in same session manually first. Any help/pointers to get that working? that way 'surveillance delete' would also be covered.

src/test/java/gov/healthit/chpl/aqa/asserts/DpmanagementUploadPageAsserts.java
ashwinimore: I'll try this

src/test/resources/Features/OCD-2797-surveillanceUpload.feature
```diff
@@ -0,0 +1,13 @@
+@Regression
+Feature: OCD-2797 - Verify surveillance upload/confirm workflow is functional
+
+  Scenario: Upload and confirm surveillance activity
+    Given I'm logged in as "ROLE_ACB"
+    And I am on Upload Products & Surveillance page
+    When I upload a surveillance activity for listing with CHPL ID 15.05.05.2760.ISCD.01.00.1.181101
```
andlar: What if this was
```suggestion
    When I upload the "SurveillanceUpload_SLI.csv" surveillance activity
```
and you use the variable to choose between the surveillances that are available? Then, when/if we need a different surveillance for some reason you can re-use the step and add a new csv to the project?

src/test/java/gov/healthit/chpl/aqa/asserts/DpmanagementUploadPageAsserts.java
andlar: Is there something in the surveillance workflow you could edit during the upload, like we do with the listing's Product & Version codes? The "number of sites surveilled" could be any integer, so if you used the day of the year (1-365) it'd likely never duplicate, as we do delete these surveillances every time the STG plan runs. Could also use one of the dates, though that restricts us to once per day

kekey1: I'm confused about what is cached regarding pending surveillance. The API call to get pending surveillance is not cached so I think any time you view the list of pending surveillance it should return a fresh list. Could you just click all the checkboxes next to each pending surveillance on the page and then the Reject All button to remove them after upload?

ashwinimore: not the pending surveillance page that's cached, manageSurveillance page where you could load surveillance for a listing and update/delete

andlar (approved):

kekey1 (approved):

PR #127, opened 2019-04-17
test: add merge version scenarios

iyerashwin14:


src/test/java/gov/healthit/chpl/aqa/asserts/ViewVersionPageAsserts.java
```diff
@@ -40,5 +40,26 @@ public void iSeeMergeLinkToMergeTheVersion(final String versionId) {
...
         assertTrue(true);
     }
+
+    /**
+     * Assert that "At least one other Version must be selected to merge" is displayed in versions to merge section.
+     * @param versionsToMergeMessage is a section in merge versions
+     */
+    @Then("^I see \"([^\"]*)\" in versions to merge section$")
+    public void iSeeMessageInVersionsToMergeSection(final String versionsToMergeMessage) {
+        ViewVersionPage.versionsToMergeMessage(getDriver()).isDisplayed();
```
andlar: This string isn't being passed into the page object anywhere, so even if this does work, changing the string in the feature file, or changing the message on the page itself, wouldn't cause this assert to fail, which means the test isn't right

src/test/java/gov/healthit/chpl/aqa/asserts/ViewVersionPageAsserts.java
```diff
@@ -40,5 +40,26 @@ public void iSeeMergeLinkToMergeTheVersion(final String versionId) {
...
+        assertTrue(true);
```
andlar: This is wrong. If anything, it should be something like
```suggestion
        assertTrue(PageObject.findMessage(theStringVariable)).isDisplayed());
```
I don't know if that'd work, but it'd be closer

src/test/resources/Features/OCD-2806-SimplifiedEditingOfVersion.feature
```diff
@@ -30,4 +30,17 @@ Scenario Outline: User logs in as ROLE and edits a version
...
+    When I navigate to the version edit page for version "54" of product "236" of developer "142"
+    And I click merge version link for version "54"
+    Then I see "At least one other Version must be selected to merge" in versions to merge section
+
+Scenario: Verify error message is displayed when the version field is left blank
+    Given I'm logged in as "ROLE_ONC"
+    When I navigate to the version edit page for version "54" of product "236" of developer "142"
+    And I click merge version link for version "54"
+    And I leave the version field blank 
+    Then I see error message "Field is required" for version "54"
```
andlar: If you put a new line at the end of the file, then adding to the bottom of the file won't show that the previous last line changed. See how line 33 of this file is marked as changed? It wasn't, it just had a newline added to it

src/test/resources/Features/OCD-2806-SimplifiedEditingOfVersion.feature
```diff
@@ -30,4 +30,17 @@ Scenario Outline: User logs in as ROLE and edits a version
...
+Scenario: Verify message is displayed in versions to merge section
+    Given I'm logged in as "ROLE_ONC"
+    When I navigate to the version edit page for version "54" of product "236" of developer "142"
+    And I click merge version link for version "54"
+    Then I see "At least one other Version must be selected to merge" in versions to merge section
+
+Scenario: Verify error message is displayed when the version field is left blank
+    Given I'm logged in as "ROLE_ONC"
+    When I navigate to the version edit page for version "54" of product "236" of developer "142"
+    And I click merge version link for version "54"
```
andlar: These steps are supposed to be friendly to non-coders, so writing them as close to real English is preferred:
```suggestion
    And I click the merge version link for version "54"
```

src/test/resources/Features/OCD-2806-SimplifiedEditingOfVersion.feature
```diff
@@ -30,4 +30,17 @@ Scenario Outline: User logs in as ROLE and edits a version
...
+    And I leave the version field blank
```
andlar: This step is mis-described. You're not leaving the field blank; if you were, there wouldn't be an action in the Step. This could be "I clear the version field", but if you really want it to be useful make it something that we can re-use, like `I set the Version field to ""`, and then, later, if we needed to actually set it to something we'd already have the Step defined

andlar (changes requested):

andlar (approved):

ashwinimore (approved):

PR #129, opened 2019-04-19
test: simplified editing of product

iyerashwin14:


src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ViewProductSteps.java
```diff
@@ -0,0 +1,61 @@
...
+     * Click Edit link of a product.
+     * @param productId is the product version id passed in page objects
+     **/
+    @And("^I click the edit link for product \"(.*)\"$")
+    public void clickEditLinkOfProduct(final String productId) {
+        ViewProductPage.editProductLink(getDriver(), productId).click();
+    }
+
+    /**
+     * Clear product name field.
```
andlar: JavaDoc is incorrect

src/test/java/gov/healthit/chpl/aqa/asserts/ViewProductPageAsserts.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Assert that product merge link exists on view version page.
+     * @param productId is the product id passed in page objects
+     */
+    @Then("^I see product merge link for product \"([^\"]*)\"$")
+    public void iSeeProductMergeLink(final String productId) {
+        ViewProductPage.mergeProductLink(getDriver(), productId).isDisplayed();
+        assertTrue(true);
```
andlar: This was something you'd changed in the Version page, right? Should do the same thing here.

`assertTrue(true)` will always pass

Same with the other three

andlar (changes requested):
I started from the bottom up, and it was almost perfect. Only when I got to the asserts did I see anything untoward. Good work in general though!

src/test/java/gov/healthit/chpl/aqa/asserts/ViewProductPageAsserts.java
iyerashwin14: Nope, I verified the text in the version feature whereas for the link it's the same as above!

andlar: So if you edit the scenario for the merge to include ROLE_ACB, and run the test. Does it pass or fail?

iyerashwin14: It fails!

andlar: Does it fail because of a stackdump, or does it write out something like "expected true, was false"?

iyerashwin14: It throws an exception and does not write anything like "expected true...."

andlar: If the exception is on the line above, then that makes sense. Taking an exception as a "good thing" to indicate a test that doesn't pass isn't good practice. That means your code isn't getting to the `assertTrue` line, and that means you're not actually testing what you want to test.

See first paragraph in this answer: https://stackoverflow.com/a/2496577/16593 It's a java based thing, but the differentiation between test **Failure** and test **Exception** is the important thing

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ViewProductSteps.java
```diff
@@ -0,0 +1,61 @@
...
+     * Click Edit link of a product.
+     * @param productId is the product version id passed in page objects
+     **/
+    @And("^I click the edit link for product \"(.*)\"$")
+    public void clickEditLinkOfProduct(final String productId) {
+        ViewProductPage.editProductLink(getDriver(), productId).click();
+    }
+
+    /**
+     * Set the product name field to "".
```
andlar: This still isn't right. You're setting the product name to the value passed in, not the empty string. It happens to be true that right now you're only passing in the empty string, but the description of the code doesn't match what it does

src/test/java/gov/healthit/chpl/aqa/asserts/ViewProductPageAsserts.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Assert that product name is correct.
+     * @param productId is the product id passed in page objects
+     */
+    @Then("^I see the edited product information is recorded and updated on View Product page for product \"(.*)\"$")
+    public void editedInformationRecordedAndUpdated(final String productId) {
+        String actualProductVersion = ViewProductPage.viewProductName(getDriver(), productId).getText();
+        ((JavascriptExecutor) getDriver()).executeScript("arguments[0];", actualProductVersion);
```
andlar: What is this doing?

src/test/java/gov/healthit/chpl/aqa/asserts/ViewProductPageAsserts.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Assert that product split link exists on view version page.
+     * @param productId is the product id passed in page objects
+     */
+    @Then("^I see product split link for product \"([^\"]*)\"$")
+    public void iSeeProductSplitLink(final String productId) {
+        try {
+            ViewProductPage.splitProductLink(getDriver(), productId).isDisplayed();
+            assertTrue(true);
```
andlar: If you put the previous line inside the assertTrue then this would make sense

src/test/java/gov/healthit/chpl/aqa/asserts/ViewProductPageAsserts.java
```diff
@@ -0,0 +1,83 @@
...
+        } catch (NoSuchElementException e) {
+            assertTrue("Unable to find product split link", false);
```
andlar: If you're specifically trying to fail the test here, use the Assert.fail method

src/test/java/gov/healthit/chpl/aqa/asserts/ViewProductPageAsserts.java
```diff
@@ -0,0 +1,83 @@
+package gov.healthit.chpl.aqa.asserts;
+
+import static org.junit.Assert.assertTrue;
+import static org.testng.Assert.assertEquals;
```
andlar, line 3: Can we settle on one or the other of these two packages? I don't care which one, but the expected/actual orders are different in their respective assertEquals, and we shouldn't be using two packages that do basically the same thing. @ashwinimore , do you have a preference?

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/asserts/ViewProductPageAsserts.java
iyerashwin14: I would prefer assertTrue

andlar: Sorry, I meant the use of the junit and testng packages. They both have "assertTrue" and "assertEquals", but we shouldn't be using both junit & testng in the same project

ashwinimore: I have been using testng package for asserts, I'd go with testng

src/test/java/gov/healthit/chpl/aqa/asserts/ViewProductPageAsserts.java
ashwinimore: or assertTrue((ViewProductPage.splitProductLink(getDriver(), productId).isDisplayed()));

src/test/java/gov/healthit/chpl/aqa/asserts/ViewProductPageAsserts.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Assert that product merge link exists on view version page.
+     * @param productId is the product id passed in page objects
+     */
+    @Then("^I see product merge link for product \"([^\"]*)\"$")
+    public void iSeeProductMergeLink(final String productId) {
+        try {
+        assertTrue(ViewProductPage.mergeProductLink(getDriver(), productId).isDisplayed());
```
andlar: Your indentation here is off. I know it's a small thing, but checkstyle should catch these too. Recommend running a checkstyle report and verifying there aren't any real issues

ashwinimore (approved):

andlar (approved):

PR #132, opened 2019-04-26
test: OCD-2823-Check that definition files exist on overnight checks

ashwinimore:


andlar (approved):
I was going to request that we not log into PROD as any user, but then scrolled back to where the examples were, and noticed you'd already not done that. Outstanding!

Approved! 🎇🎆🎉

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/LoginSteps.java
```diff
@@ -80,5 +80,66 @@ public void logInAsRole(final String role) {
         LoginPage.loginButton(getDriver()).click();
         getWait().until(ExpectedConditions.visibilityOf(LoginPage.logoutButton(getDriver())));
     }
+
```
kekey1, line 83: This function looks ALMOST like a copy of the loginAsRole method. Could you see if you could make it just a single function that will do what you want here? Like if tEnv is null it just does what loginAsRole does now?

kekey1 (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/LoginSteps.java
ashwinimore: Yes, it is almost identical, with tEnv being passed as a parameter is the only difference. I did think of modifying and reusing existing step instead of adding new one, but then realized existing step is being used in 14 other features where tEnv is not needed and editing that step would mean editing it in all those features and step definitions and also adding test data for env. in each of those scenarios where login step is used. We already have url set as property in case we want to run tests in other environment so editing existing step and adding test data each time for a field that is not needed seemed not so useful to me. Instead, i thought if this step is added, for scenarios like these where we want to run tests in all environments, we can have a choice of using a specific login step.

kekey1: I guess I'm just trying to avoid having code that's almost identical in two places. That's a recipe for forgetting to change one of the places. Can you refactor the two methods in some way so that maybe most of the shared code is in a single private method that they both call?

kekey1 (approved):

PR #136, opened 2019-05-02
test: simplified editing of developer

iyerashwin14:


src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ViewDeveloperSteps.java
```diff
@@ -0,0 +1,80 @@
...
+        wait.until(ExpectedConditions.visibilityOf(ViewDeveloperPage.viewDeveloperPageTitle(getDriver())));
+    }
+
+    /**
+     * Edit a developer.
+     * @param developerId is the developer id passed in page objects
+     **/
+    @And("^I edit for developer \"(.*)\"$")
+    public void editDeveloper(final String developerId) {
+        ViewDeveloperPage.editDeveloperLink(getDriver(), developerId).click();
```
kekey1: Could you call the below method clickEditLinkOfDeveloper(developerId) here instead of writing this line again?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ViewDeveloperSteps.java
```diff
@@ -0,0 +1,80 @@
...
+        ViewDeveloperPage.developerNameField(getDriver()).clear();
+        ViewDeveloperPage.developerNameField(getDriver()).sendKeys(getCurrentDate());
```
kekey1: Could you call the below method setDeveloperNameField() instead of the clear and sendKeys methods called above?

src/test/resources/Features/OCD-2873-SimplifiedEditingOfDeveloper.feature
```diff
@@ -0,0 +1,64 @@
...
+
+Scenario Outline: User logs in as ROLE and sees developer edit link exists on new developer page
+    Given I'm logged in as "<ROLE>"
+    When I navigate to the developer edit page for developer "1854"
+    Then I see developer edit link for developer "1854"
+    Examples:
+      |ROLE       |
+      |ROLE_ACB   |
+      |ROLE_ADMIN |
+      |ROLE_ONC   |
```
kekey1, line 13: Could we also have a scenario for a user who is not logged in at all and test that they don't see the edit/merge/split links?

src/test/resources/Features/OCD-2873-SimplifiedEditingOfDeveloper.feature
```diff
@@ -0,0 +1,64 @@
...
+    Examples:
+      |ROLE       |
+      |ROLE_ADMIN |
+      |ROLE_ONC   |
+
+Scenario Outline: User logs in as ROLE and edits a developer
+    Given I'm logged in as "<ROLE>"
+    When I navigate to the developer edit page for developer "<developer id>"
+    And I edit for developer "<developer id>"
+    Then I see the edited developer information is recorded and updated on View Developer page
```
kekey1, line 38: I feel like it would be better to specify here in the feature what the developer name should be edited to and check against that value rather than using the getCurrentDate() method to generate a name.

kekey1 (changes requested):

src/test/resources/Features/OCD-2873-SimplifiedEditingOfDeveloper.feature
andlar: This is one of the paradigms we have for automated tests. We can't set the name to a specific value, because if we do that, and it works "today", when the tests run "tomorrow" we won't know if we actually change the name tomorrow or not, as it won't be a change. We do something similar in the "upload" tests, where the Product and Version Codes get set to values dependent on the date. I don't like the _format_ of the name that's being used, but it's functional

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ViewDeveloperSteps.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Class ViewDeveloperSteps definition.
+ */
+public class ViewDeveloperSteps extends Base {
+
+    /**
+     * Get user to the View Developer Page.
+     * @param developerId is Developer ID
+     **/
+    @Given("^I navigate to the developer edit page for developer \"([^\"]*)\"$")
```
andlar: This isn't navigating to the edit page. This is navigating to the developer page. If you want to navigate to the edit page then the URL would be `#/organizations/developers/{ developerID }/edit`. Should either change the step definition or change the URL

src/test/java/gov/healthit/chpl/aqa/asserts/ViewDeveloperPageAsserts.java
```diff
@@ -0,0 +1,102 @@
...
+            fail("Unable to find developer merge link");
+        }
+    }
+
+    /**
+     * Assert that developer name is correct.
+     */
+    @Then("^I see the edited developer information is recorded and updated on View Developer page$")
+    public void editedDeveloperInformationRecordedAndUpdated() {
+        String actualDeveloperVersion = ViewDeveloperPage.viewDeveloperName(getDriver()).getText();
```
andlar: What is `Version` doing here? Does this assert have anything to do with Versions?

andlar (changes requested):

src/test/resources/Features/OCD-2873-SimplifiedEditingOfDeveloper.feature
kekey1: Oh.. hmm.. when I was looking at ACB names a while ago it looked to me like in those tests we rename the ACB to a certain value and then later on in the test rename it back to something else. Also, larger issue, but I feel like we should always be starting a test suite from a known state of data - meaning we should run the bamboo job before the automated tests run. Otherwise how could we EVER assume that any acb, developer, whatever had a certain name? I could log in as an admin and rename the developer during the workday to be today's date and then you'd have the same problem you described above - how would you know if the edit worked?

andlar: In the ACB rename one we'd done that all in the same step, including the asserts. That isn't great as it combines the action with the assert, but was the only way we could undo the things. And we needed to undo things on an ACB rename because otherwise all of the other tests using upload / filter / etc. failed.

With Developers, if a name changes here or there, unless we're using that Developer in another test it won't have any lasting impact. I'm up for ideas about undoing things in general, but we don't have that really set up right now. Maybe have jenkins do a database dump before running the tests, and then a restore after? Would definitely slow down the system, and might need to stop tomcat at the same time, which is complicated.

As for the "known state", I agree, but the hard part is that our users are messing with our state on a daily basis. For that reason I consider most of these tests as slightly fragile, but it's an issue we can't really do much about, outside of setting up an environment that doesn't have end user input

src/test/resources/Features/OCD-2873-SimplifiedEditingOfDeveloper.feature
```diff
@@ -0,0 +1,64 @@
...
+    And I click the edit link for developer "1854"
+    And I set the email field to ""
+    Then I see "Field is required" error message for email field
+
+Scenario: Verify message is displayed in developers to merge section
+    Given I'm logged in as "ROLE_ONC"
+    When I navigate to the developer edit page for developer "1854"
+    And I click the merge developer link for developer "1854"
+    Then I see "At least one other Developer must be selected to merge" in developers to merge section
+
```
ashwinimore, line 64: Add a similar scenario for developer split to verify message "At least one Product must be selected to move"

ashwinimore: Also recommend adding a scenario to login as ACB and setting Transparency Attestation for logged in ACB

ashwinimore (changes requested):

andlar (approved):

src/test/java/gov/healthit/chpl/aqa/asserts/ViewDeveloperPageAsserts.java
```diff
@@ -109,5 +109,43 @@ public void iSeeMessageInProductsMovingToNewDeveloperSection(final String expect
...
+    }
+
+    /**
+     * Assert that developer merge link does not exist on view developer page.
+     * @param developerId is the developer id passed in page objects
+     */
+    @Then("^I do not see developer merge link for developer \"([^\"]*)\"$")
+    public void iDoNotSeeDeveloperMergeLink(final String developerId) {
+        try {
+            assertTrue(!ViewDeveloperPage.mergeDeveloperLink(getDriver(), developerId).isDisplayed());
```
andlar: Doesn't this throw an exception if it can't find the element? If that's the case, then this test won't ever pass, as it'll go to the catch block and not show a passed assert

src/test/java/gov/healthit/chpl/aqa/asserts/ViewDeveloperPageAsserts.java
```diff
@@ -0,0 +1,154 @@
...
+     * Assert that developer edit link does not exist on view developer page.
+     * @param developerId is the developer id passed in page objects
+     */
+    @Then("^I do not see developer edit link for developer \"([^\"]*)\"$")
+    public void iDoNotSeeDeveloperEditLink(final String developerId) {
+        boolean developerEditLinkFound = false;
+        try {
+            developerEditLinkFound = ViewDeveloperPage.editDeveloperLink(getDriver(), developerId).isDisplayed();
+        } catch (NoSuchElementException e) {
+        }
```
andlar, line 126: In this assert, you're expecting an exception to be thrown, right? If that's the case, I'd put the assert inside the catch block. Empty catch blocks are a code smell

iyerashwin14: Okay, so if I'd put the assert inside the catch block the test case would never fail. Assert needs to be outside the catch block. Currently, after an exception is been thrown it executes the assert statement. 

Consider the scenario like below:-
Given I'm logged in as "ROLE_ONC"
When I navigate to the developer page for developer "1854"
Then I do not see developer edit link for developer "1854"

The test case fails because if I login - I'm able to see the developer edit link but if I have the assert statement inside the catch block it does not fail.

So instead of leaving the catch block empty I could print something like:- 
catch (NoSuchElementException e) {
            System.out.println("Developer split link is not available");
        }

iyerashwin14: Please let me know if I'm correct

andlar: In the try block, you're trying to find an element. Is that element found? If it is, then in that try block you should have something like `fail("Found element when shouldn't have")`. If it's not found, does it throw an exception? If it does, and that's what we _want_ to have happen, then you should catch that exception, and in the catch block put your `assertTrue(!developerEditLinkFound, "Error, developer edit link is available);`.

If not finding an element doesn't throw an exception, then there shouldn't be a try/catch block at all

kekey1 (approved):

src/test/java/gov/healthit/chpl/aqa/asserts/ViewDeveloperPageAsserts.java
```diff
@@ -146,9 +148,10 @@ public void iDoNotSeeDeveloperMergeLink(final String developerId) {
         boolean developerMergeLinkFound = false;
         try {
             developerMergeLinkFound = ViewDeveloperPage.mergeDeveloperLink(getDriver(), developerId).isDisplayed();
+            fail("Found element when shouldn't have");
         } catch (NoSuchElementException e) {
+            assertTrue(!developerMergeLinkFound, "Developer merge link is available");
```
andlar: Is there an `assertFalse`? If so, it would be more clear to use that rather than the `!`.

Oh, one other thing. Have you tried adding a "log in" step in this feature temporarily, and seen that the test actually does fail? It's always a good idea to try to make sure your test will actually fail sometimes, otherwise you're can't be sure you're exercising the system correctly

iyerashwin14: Yes, I tried adding a "log in" step and verified that the test fails.

ashwinimore (approved):

PR #138, opened 2019-05-07
test: add developer, product and version name change to the eye

iyerashwin14:


src/test/resources/Features/OCD-2049-productHistoryEye.feature
```diff
@@ -1,5 +1,5 @@
 Feature: OCD-2049 - Verify 'Eye' on listing details page works and displays Product History 
-
+         OCD-2822 - Add Developer, Product and Version name change to the eye
```
andlar: If this feature file now has two tickets in it, the file name should change to indicate that it's not specific to a single feature

src/test/java/gov/healthit/chpl/aqa/asserts/ProductHistoryEyeAsserts.java
```diff
@@ -0,0 +1,45 @@
...
+public class ProductHistoryEyeAsserts extends Base {
+
+    /**
+     * Assert that expected developer edit details is displayed on product history eye.
+     * @param expectedDeveloperEditDetails is "Developer changed from"
+     */
+    @Then("^I see the product history of developer edit is recorded as \"([^\"]*)\"$")
+    public void iSeeProductHistoryOfDeveloper(final String expectedDeveloperEditDetails) {
+        String actualDeveloperEditDetails = ProductHistoryEyePage.productHistoryDeveloper(getDriver()).getText();
+        assertTrue((actualDeveloperEditDetails.startsWith(expectedDeveloperEditDetails)));
```
kekey1: I guess I'm not sure about this logic... you're looking up the element in ProductHistoryEyePage by looking for the text "Developer changed from". But then you are asserting that the elements text starts with "Developer changed from". This seems like a duplicate check to me. If the ProductHistoryEyePage.productHistoryDeveloper() method returns an element at all then I think you know that the edit history is there. It's not really wrong but just seems like you don't need to check twice.

kekey1 (approved):

src/test/java/gov/healthit/chpl/aqa/pageObjects/ProductHistoryEyePage.java
```diff
@@ -0,0 +1,27 @@
+package gov.healthit.chpl.aqa.pageObjects;
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+/**
+ * Class ProductHistoryEyePage definition.
+ */
+public final class ProductHistoryEyePage {
```
andlar: The URL for where this page is now uses /listing, and not product. Recommend changing to use Listing instead of Product here, that way in the WebElement productHistoryProduct you can make it look reasonable by using listingHistoryProduct

src/test/java/gov/healthit/chpl/aqa/asserts/ProductHistoryEyeAsserts.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * Assert that expected product edit details is displayed on product history eye.
+     * @param expectedProductEditDetails is "Product changed from"
+     */
+    @Then("^I see the product history of product edit is recorded as \"([^\"]*)\"$")
+    public void iSeeProductHistoryOfProduct(final String expectedProductEditDetails) {
+        String actualProductEditDetails = ProductHistoryEyePage.productHistoryProduct(getDriver()).getText();
+        assertTrue((actualProductEditDetails.startsWith(expectedProductEditDetails)));
```
andlar: Shouldn't use the text you're looking for as the data to use to find a PageObject. If the Page changes, the PageObject should be the only thing that needs to change, but here you'd need to change the assert as well. Also, if this test works the first time, but fails to do the edit the next day, it'll still find that history in the eye, when it should report a failure

andlar (changes requested):
I think the test idea here isn't right. I think the final "Then" in the Feature should be something like `Then I see the listing history eye shows a Product name change on today's date`. In your page object model, you should return a list of elements, or the table that holds them, with all of the changes, then, in the assert step, you can loop through those items looking for both a change of the Product name (or Version/Developer) that happened on the same date (or maybe "within the last n minutes")

src/test/java/gov/healthit/chpl/aqa/asserts/ProductHistoryEyeAsserts.java
ashwinimore: I think only asserting partial text isn't enough as there might be existing history on product that includes that text or like Andrew said above, from previous day's edit. A good check would be to combine the text assert with date validation in same row in 'date' column  when activity happened

ashwinimore (changes requested):

src/test/java/gov/healthit/chpl/aqa/asserts/ListingHistoryEyeAsserts.java
```diff
@@ -0,0 +1,64 @@
...
+public class ListingHistoryEyeAsserts extends Base {
+
+    /**
+     * Assert that expected product edit details is displayed on product history eye.
+     * @param expectedEditDetails is "Developer/Product/Version change"
+     * @throws Exception of type ParseException or InterruptedException
+     */
+    @Then("^I see the listing history eye shows a \"([^\"]*)\" on today's date$")
+    public void iSeeListingHistory(final String expectedEditDetails) throws Exception {
+        Thread.sleep(SLEEP_TIME);
```
iyerashwin14, line 29: History table takes a long time to load - thread.sleep does not look good but I tried different options and nothing worked for me.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Base.java
```diff
@@ -162,5 +164,24 @@ public String getCurrentDate() {
...
+     * @param date in MMMM dd,yyyy with time stamp and GMT time zone
+     * @throws ParseException if unable to parse the date
+     * @return date
+     */
+    public Date getFormatedDateTimeZone(final String date) throws ParseException {
+        Date dateInstance;
+        DateFormat dateFormat = new SimpleDateFormat("MMMM dd,yyyy hh:mm:ss a z");
+        dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
+        if (date == null) {
+            String currentDate = dateFormat.format(new Date());
```
andlar:
```suggestion
            return new Date();
```

There's no reason to create a string using a new Date only to then parse it into a Date object. Just return the new Date.

Should also probably move this entire check to the first line of the method. No reason to do any other work if the Date doesn't exist in the first place

src/test/java/gov/healthit/chpl/aqa/asserts/ListingHistoryEyeAsserts.java
```diff
@@ -0,0 +1,64 @@
...
+import gov.healthit.chpl.aqa.pageObjects.ListingHistoryEyePage;
+import gov.healthit.chpl.aqa.stepDefinitions.Base;
+
+/**
+ * Class ListingHistoryEyeAsserts definition.
+ */
+public class ListingHistoryEyeAsserts extends Base {
+
+    /**
+     * Assert that expected product edit details is displayed on product history eye.
```
andlar: Comment says "expected product edit" when it could be any of three types of edit. Also says "product history eye" when we should be using "listing history eye" to ~avoid~ reduce confusion

src/test/java/gov/healthit/chpl/aqa/asserts/ListingHistoryEyeAsserts.java
```diff
@@ -0,0 +1,64 @@
...
+        ListingHistoryData listingHistoryDataInst = null;
+        String date = null;
+        for (String activity : listingHistoryDetails) {
+            if ((activityCount % 2) == 0) {
+                date = activity;
+            } else {
+                if (activity.startsWith(expectedEditDetails)) {
+                    listingHistoryDataInst = new ListingHistoryData();
+                    listingHistoryDataInst.setDate(date);
+                    listingHistoryDataInst.setActivity(activity);
```
andlar: You never use this Activity data later. I don't think adding the model to this helps in any real way

src/test/java/gov/healthit/chpl/aqa/asserts/ListingHistoryEyeAsserts.java
```diff
@@ -0,0 +1,64 @@
...
+                    break;
```
andlar: The way I read this, the loop exits as soon as it finds any activity that starts with the "expectedEditDetails" value. What happens if the table is sorted backwards?

src/test/java/gov/healthit/chpl/aqa/asserts/ListingHistoryEyeAsserts.java
```diff
@@ -0,0 +1,64 @@
...
+                }
+            }
+            activityCount++;
+        }
+        if (listingHistoryDataInst == null) {
+            assertFalse("No edit activity recorded for today", true);
```
andlar: `true` will always be `true`, so asserting that it's `false` doesn't make sense. If you're trying to immediately fail, use the `fail` method: https://static.javadoc.io/org.testng/testng/6.8.17/org/testng/Assert.html#fail(java.lang.String)

src/test/java/gov/healthit/chpl/aqa/asserts/ListingHistoryEyeAsserts.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+        Calendar calInstanceExpectedDate = Calendar.getInstance();
+        Calendar calInstanceCurrentDate = Calendar.getInstance();
+        Date actualDate = getFormatedDateTimeZone(listingHistoryDataInst.getDate());
+        Date currentDate = getFormatedDateTimeZone(null);
```
andlar: Why create a Calendar for "current", then immediately convert it back to a Date? Why not just create a Date using new Date()?

src/test/java/gov/healthit/chpl/aqa/asserts/ListingHistoryEyeAsserts.java
```diff
@@ -0,0 +1,64 @@
+package gov.healthit.chpl.aqa.asserts;
+
+import static org.junit.Assert.assertFalse;
```
andlar: Wrong `assertFalse`; shouldn't use junit

src/test/java/gov/healthit/chpl/aqa/asserts/ListingHistoryEyeAsserts.java
```diff
@@ -0,0 +1,64 @@
...
+        if (listingHistoryDataInst == null) {
+            assertFalse("No edit activity recorded for today", true);
+        }
+        Calendar calInstanceExpectedDate = Calendar.getInstance();
+        Calendar calInstanceCurrentDate = Calendar.getInstance();
+        Date actualDate = getFormatedDateTimeZone(listingHistoryDataInst.getDate());
+        Date currentDate = getFormatedDateTimeZone(null);
+        calInstanceExpectedDate.setTime(actualDate);
+        calInstanceCurrentDate.setTime(currentDate);
+        boolean listingHistoryRecorded = false;
```
andlar: It doesn't make sense to create a boolean, then use an if to decide whether to set that boolean to true. Just set `boolean listingHistoryRecorded = (the condition in your if)`, and assert about it later

src/test/java/gov/healthit/chpl/aqa/asserts/ListingHistoryEyeAsserts.java
```diff
@@ -0,0 +1,64 @@
...
+        if ((calInstanceExpectedDate.get(Calendar.YEAR) == calInstanceCurrentDate.get(Calendar.YEAR))
```
andlar: Don't use the calendarInstance to compare things

```
Date foundDate = getFormatedThing(string from the table)
Date actualDate = new Date
if (actualDate.getTime - foundDate.getTime < some number of milliseconds) {
// then the "current time" is less than "some number of milliseconds" after the "found date stamp", and your activity happened "recently"
```

You're going from a string, to a date, to a calendar instance, and then only comparing calendars down to the day of month resolution, which means that if you ran this test "now", and it worked, and then you ran it again in a few minutes, and something broke in the update code in the meantime, your test would still pass, even though the update might not have worked

src/test/java/gov/healthit/chpl/aqa/model/ListingHistoryData.java
```diff
@@ -0,0 +1,24 @@
+package gov.healthit.chpl.aqa.model;
+
+/**
+ * Class ListingHistoryData model.
+ */
+public class ListingHistoryData {
```
andlar: I don't think this class is needed at all

src/test/resources/Features/OCD-2049 & OCD-2822-productHistoryEye.feature
```diff
@@ -0,0 +1,70 @@
+Feature: OCD-2049 - Verify 'Eye' on listing details page works and displays Product History
```
andlar: 1. Spaces in filenames are evil, always. Windows lets you get away with it, and generally OS's handle it okay, but as a rule, never have spaces in filenames
2. Look at the other feature files that have multiple tickets in them; none of them have the ticket number in the filename

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/asserts/ListingHistoryEyeAsserts.java
iyerashwin14: In this case, I'm assuming the first row is the latest updated one. So I'm always considering the first row for which the activity is matched.

andlar: I recognize that, and that's exactly the problem. What happens if some other process updates the same Listing in-between your update and your check? What happens if, for whatever reason, we decide to change the order of these items? What happens if we end up adding a new activity type that is added when you change the name, and it shows up as more recent than the name change? This code _assumes_ the most recent item is the right one, but it doesn't need to, and by doing so it makes the test more brittle

iyerashwin14: I understand, would try to come up with a better approach. Thanks!

andlar (approved):

ashwinimore (approved):

PR #139, opened 2019-05-07
fix: minimize pop-up of login/out section on new admin-nav bar

iyerashwin14:


src/test/java/gov/healthit/chpl/aqa/stepDefinitions/LoginSteps.java
```diff
@@ -79,6 +79,9 @@ public void logInAsRole(final String role) {
         LoginPage.password(getDriver()).sendKeys(password);
         LoginPage.loginButton(getDriver()).click();
         getWait().until(ExpectedConditions.visibilityOf(LoginPage.logoutButton(getDriver())));
+        //minimize the pop up of login/out section
+        wait.until(ExpectedConditions.visibilityOf(LoginPage.loginLogoutPopUp(getDriver())));
```
andlar: Why do you need to wait for this? Isn't it already visible?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/LoginSteps.java
```diff
@@ -79,6 +79,9 @@ public void logInAsRole(final String role) {
         LoginPage.password(getDriver()).sendKeys(password);
         LoginPage.loginButton(getDriver()).click();
         getWait().until(ExpectedConditions.visibilityOf(LoginPage.logoutButton(getDriver())));
```
andlar, line 79: Why are there two different "wait"s here? One from the Base.java page, found in the "getWait()" call, and the other created in this method, on line 72. There should only be one, and it should probably be the one from the getWait method

andlar (changes requested):

kekey1 (approved):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/LoginSteps.java
ashwinimore: I think it's been there from the time before we had Base.java, but agree that getWait() should be used now that we have it and is used in other tests

ashwinimore (approved):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/LoginSteps.java
```diff
@@ -69,16 +68,18 @@ public void logInAsRole(final String role) {
...
         LoginPage.loginLogoutPopUp(getDriver()).click();
         //when it's popped up we can see the username
-        wait.until(ExpectedConditions.visibilityOf(LoginPage.username(getDriver())));
+        getWait().until(ExpectedConditions.visibilityOf(LoginPage.username(getDriver())));
         LoginPage.username(getDriver()).sendKeys(username);
         LoginPage.password(getDriver()).sendKeys(password);
         LoginPage.loginButton(getDriver()).click();
         getWait().until(ExpectedConditions.visibilityOf(LoginPage.logoutButton(getDriver())));
+        //minimize the pop up of login/out section
+        getWait().until(ExpectedConditions.visibilityOf(LoginPage.loginLogoutPopUp(getDriver())));
```
andlar, line 81: Question is still valid; there's a wait above that this element is visible. Does it disappear before getting to this wait? If so, this is fine, if not, then it shouldn't be waiting again

iyerashwin14: It disappears at times because of the new admin nav bar and throws NoSuchElementException so that's the reason I added wait here

andlar (approved):

PR #140, opened 2019-05-09
Ocd 2894-Verify order of headings, location and new nav link for Recommended Web Browsers section on overiew page

ashwinimore:


src/test/java/gov/healthit/chpl/aqa/asserts/OverviewPageAsserts.java
```diff
@@ -70,4 +71,40 @@ public void urlForNavLinksOnOverviewPage(final String url, final String linkText
       getDriver().getPageSource().contains(linkText);
       assertTrue(OverviewPage.headerText(getDriver(), headerNumber).isDisplayed());
     }
+
+    /**
+     * Asserts that headings are in the correct order on overview page.
+     */
+    @Then("^I should see the sections in the following order: Announcement, What is the CHPL, Recommended Web Browsers, "
+            + "Using the CHPL Website, ONC Certification Program, For EHR Developers, ONC-ACB and ONC-ATL Information$")
+    public void verifySectionOrder() {
```
kekey1: Can you make the section names be part of the feature file and use a List<String> variable as an argument to this method? That way they won't have to be adjusted in both the @Then and the code if they change in the future.

kekey1 (approved):

iyerashwin14 (approved):

andlar (approved):

PR #143, opened 2019-05-29
test: simplified editing of listing - basic data

iyerashwin14:


src/test/java/gov/healthit/chpl/aqa/asserts/ListingDetailsAsserts.java
```diff
@@ -4,12 +4,14 @@
...
 
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
 
 import cucumber.api.java.en.Then;
+import gov.healthit.chpl.aqa.pageObjects.DpManagementPage;
```
andlar: If you're asserting stuff on the Listing Details page, then you shouldn't nee the DpManagementPage. If you're just using the DpManagmentPage for the Toast thing, then the Toast thing should be moved to the BasePage, where it could be used on any page

andlar (changes requested):

ashwinimore (approved):

src/test/java/gov/healthit/chpl/aqa/asserts/ViewVersionPageAsserts.java
```diff
@@ -9,7 +9,7 @@
 import org.openqa.selenium.support.ui.WebDriverWait;
 
 import cucumber.api.java.en.Then;
-import gov.healthit.chpl.aqa.pageObjects.DpManagementPage;
+import gov.healthit.chpl.aqa.pageObjects.BasePage;
```
andlar: The "ViewVersionPage" page object should extend from the "BasePage" page object, which would mean that the ViewVersionPage PO would have the updateSuccessfulToastContainer as part of itself without importing the BasePage at all in the assert files

andlar (approved):

PR #152, opened 2019-06-08
test: update CMS ID test to use new nav

ashwinimore:


iyerashwin14 (approved):

kekey1 (approved):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ViewDeveloperSteps.java
```diff
@@ -21,6 +21,7 @@
      **/
     @Given("^I navigate to the developer page for developer \"([^\"]*)\"$")
     public void editDeveloperPage(final String developerId) {
+        getDriver().get(getUrl());
```
andlar, line 24: Why go to the "basic" page, then go to the developer page? And if so, why is there no "wait" to make sure you get to the basic page before going to the developer page?

ashwinimore: Accessing #/organizations/developers/developerId directly, when user is not logged in, redirects user to basic page. But if user loads basic page first and the loads #/organizations/developers/developerId, it works fine. Something with new nav I guess. I was going to chat about it but wanted to make this temp update so tests continue to run until nav for non logged user is reviewed & updated.

andlar (approved):

PR #159, opened 2019-06-21
test: update tests to use new admin nav and listing edit workflow

ashwinimore:


src/test/resources/Features/verifyDataInCriteriaDetails.feature
```diff
@@ -1,4 +1,5 @@
 @Regression
+@criteriaDetails
```
andlar, line 2: What's this for?

andlar (approved):
Curious about one thing, but approved

src/test/resources/Features/verifyDataInCriteriaDetails.feature
ashwinimore: It's much easier to run specific group of tests on Jenkins using tags, so added this one to execute tests that validate criteria details

kekey1 (approved):

PR #161, opened 2019-06-27
test: update failing manage surveillance test to use new workflow

ashwinimore:


src/test/java/gov/healthit/chpl/aqa/asserts/ManageSurveillanceAsserts.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Verify surveillance search results load by validating CHPL ID in results.
+     * @param chplId is chplId to look up
+     * @throws Exception if results did not load and chplId was not found
+     */
+    @Then("^I see the surveillance results for \"([^\"]*)\"$")
+    public void testSurveillanceResultsAsExpected(final String chplId) throws Exception {
+        try {
+            String actualString = DpManagementPage.manageProductSurveillance(getDriver()).getText();
```
andlar, line 24: Going to approve this, but this test is using the DpManagementPage, which is the page that's going to go away completely as soon as the adminNav flag goes away. It would be good to migrate any tests that we're changing / move any that we're creating, to use the new nav pages as we can

andlar (approved):

src/test/java/gov/healthit/chpl/aqa/asserts/ManageSurveillanceAsserts.java
ashwinimore: I'll start working on renaming existing/adding new pages to align with new nav soon

iyerashwin14 (approved):

PR #162, opened 2019-07-17
test: automate complaint tests

ashwinimore:


src/test/resources/Features/surveillanceComplaints.feature
```diff
@@ -0,0 +1,70 @@
...
+    And I click save button
+    Then a new complaint with given "<ONC-ACB Complaint ID>" should be added to CHPL 
+Examples:
+    |ONC-ACB|Received Date|Closed Date|ONC-ACB Complaint ID|ONC Complaint ID|Complainant Type|Summary|
+    |Drummond Group|07/12/2019|07/12/2019|ONCACB001|ONC001|Developer|Test Summary|
+    
+Scenario: Edit a complaint
+    Given I'm logged in as "ROLE_ACB"
+    And I am on complaints page
+    When I click Edit complaint for ONC-ACB Complaint ID "ONCACB001"
```
andlar: This step assumes the previous one has already run. I don't have a better option for this right now; we need a "real" complaint to edit, so will not block this PR based on this, but want to update it once we actually have real data

src/test/resources/Features/surveillanceComplaints.feature
```diff
@@ -0,0 +1,70 @@
...
+    Then "Edit Complaint" form should open
+    When I edit data in ONC Complaint ID field to "TsT0NC-1525"
+    And I click save button on Edit complaint form
+    Then the edits should save and show "TsT0NC-1525" in complaints view table
+    
+Scenario: Delete a complaint
```
andlar, line 33: This is even harder. We can't assume a complaint exists. I wonder if this scenario needs to create a complaint and then delete it...

src/test/resources/Features/surveillanceComplaints.feature
```diff
@@ -0,0 +1,70 @@
...
+    Given I'm logged in as "ROLE_ACB"
+    And I am on complaints page
+    And I click Add New Complaint link
+    And I enter complaint data in fields: "SLI Compliance" "07/12/2019" "07/12/2019" "ONCACB002" "ONC001" "Developer" "TBD"
+    When I associate listing "15.05.05.2760.ISAD.01.00.0.180412" to the complaint
+    And I click save button
+    And I click Edit complaint for ONC-ACB Complaint ID "ONCACB002"
+    Then the listing "15.05.05.2760.ISAD.01.00.0.180412" should be associated to the complaint
+    And I click delete button to delete selected complaint
+    And I confirm delete
```
andlar, line 53: Is this part of the scenario? Why is it deleting things?

src/test/resources/Features/surveillanceComplaints.feature
```diff
@@ -0,0 +1,70 @@
...
+Scenario: Associate a criteria to a complaint
+    Given I'm logged in as "ROLE_ACB"
+    And I am on complaints page
+    And I click Edit complaint for ONC-ACB Complaint ID "ONCACB003"
+    When I associate criteria "170.315 (a)(1)" to the complaint
+    And I click save button on Edit complaint form
+    And I click Edit complaint for ONC-ACB Complaint ID "ONCACB003"
+    Then the criteria "170.315 (a)(1)" should be associated to the complaint
+    And I click delete button to delete selected complaint
+    And I confirm delete
```
andlar: Same question re: deletes

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceComplaintsSteps.java
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Edit a complaint.
+     * @param acbComplaintId is a ACB complaint ID of a complaint to be edited
+     */
+    @When("^I click Edit complaint for ONC-ACB Complaint ID \"([^\"]*)\"$")
+    public void clickEditComplaint(final String acbComplaintId) {
+        getDriver().navigate().refresh();
+        getWait().until(ExpectedConditions.visibilityOf(SurveillanceComplaintsPage.complaintsTable(getDriver())));
+
+        List<WebElement> acbCmomplaintIdValue = getDriver().findElements(By.xpath("//*[@id=\"main-content\"]/div/div/div/div/table/tbody/tr/td[4]"));
```
andlar: Suggest requesting an id or class or something for these; this xpath is way too deep

...later

Wait, why is a Step so concerned with xpaths? Can this be a Page Element that the step checks?

andlar (approved):
No required changes; some questions. Don't see anything that requires change, but do have some questions. Will approve as it stands

src/test/resources/Features/surveillanceComplaints.feature
ashwinimore: I am deleting the complaint that was added, so the test complaints just don't pile up, add to the page load time and next time aqa tests run, there isn't existing complaint with that ID. I was thinking of using dates for IDs but then thought this could be a good alternative.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceComplaintsSteps.java
ashwinimore: I was going to ask for Ids, but i had worked on this part on weekend, so i used what was there. I'll request a better locator. I tried adding it as a page element and using with list and it wouldn't work at that time, but was running fine with xpaths, not sure why - selenium glitch maybe.

src/test/resources/Features/surveillanceComplaints.feature
ashwinimore: I thought about adding a new complaint and then doing some operation on it but the also thought it probably wasn't necessary. I agree with you that there shouldn't be dependency between scenarios. I'll make that change for this & delete scenario (next comment).

tmy1313 (approved):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceComplaintsSteps.java
```diff
@@ -0,0 +1,139 @@
...
+            final String acbComplaintId, final String oncComplaintId, final String complainantType,
+            final String summary) {
+        SurveillanceComplaintsPage.oncACBField(getDriver()).sendKeys(acb);
+        SurveillanceComplaintsPage.receivedDate(getDriver()).sendKeys(receivedDate);
+        SurveillanceComplaintsPage.closedDate(getDriver()).sendKeys(closedDate);
+        SurveillanceComplaintsPage.oncACBComplaintId(getDriver()).sendKeys(acbComplaintId);
+        SurveillanceComplaintsPage.oncComplaintId(getDriver()).sendKeys(oncComplaintId);
+        SurveillanceComplaintsPage.complainantType(getDriver()).sendKeys(complainantType);
+        SurveillanceComplaintsPage.summary(getDriver()).sendKeys(summary);
+        ((JavascriptExecutor) getDriver()).executeScript("arguments[0].click();", SurveillanceComplaintsPage.complainantContactedCheckbox(getDriver()));
```
andlar: I see that you're checking all of these checkboxes, but there was no boolean in the arguments as to whether they should be selected or not. I think I'd prefer not selecting these at all unless there is an explicit request to do so as an argument

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceComplaintsSteps.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Edit a complaint.
+     * @param acbComplaintId is a ACB complaint ID of a complaint to be edited
+     */
+    @When("^I click Edit complaint for ONC-ACB Complaint ID \"([^\"]*)\"$")
+    public void clickEditComplaint(final String acbComplaintId) {
+        getDriver().navigate().refresh();
+        getWait().until(ExpectedConditions.visibilityOf(SurveillanceComplaintsPage.complaintsTable(getDriver())));
+
+        List<WebElement> acbComplaintIdValue = getDriver().findElements(By.xpath("//*[@id=\"main-content\"]/div/div/div/div/table/tbody/tr/td[4]"));
```
andlar, line 69: Shouldn't XPaths be in the Page Object files, not in the Steps?

src/test/java/gov/healthit/chpl/aqa/pageObjects/SurveillanceComplaintsPage.java
```diff
@@ -0,0 +1,117 @@
...
+    public static WebElement deleteComplaintButton(final WebDriver driver) {
+        return driver.findElement(By.cssSelector("a[title='Delete Complaint']"));
+    }
+
+     public static WebElement developerContactedCheckbox(final WebDriver driver) {
+        return driver.findElement(By.id("developer-contacted"));
+    }
+
+    public static WebElement editComplaintButton(final WebDriver driver) {
+        return driver.findElement(By.xpath("//*[@id=\"main-content\"]/div/div/div/div[3]/table/tbody/tr[1]/td[7]/button[1]/i"));
```
andlar: Recommend requesting some sort of ID for this, though not required

src/test/java/gov/healthit/chpl/aqa/pageObjects/SurveillanceComplaintsPage.java
```diff
@@ -0,0 +1,117 @@
...
+    public static WebElement addNewComplaintButton(final WebDriver driver) {
+        return driver.findElement(By.id("add-new-complaint"));
+    }
+
+    public static WebElement addComplaintFormTitle(final WebDriver driver) {
+        return driver.findElement(By.xpath("//*[@id=\"chpl-complaint\"]/div[1]/h2"));
+    }
+
+    public static WebElement associatedCriteriaValue(final WebDriver driver) {
+        return driver.findElement(By.xpath("//*[@id=\"chpl-complaint\"]/div[13]/div[1]/span/button"));
```
andlar, line 26: One of the comments on this ticket has some rearrangement requests. My guess is that the `div[13]` is not going to work for long. Maybe each of the form elements could have an ID that you could start from? Or the button should...

src/test/java/gov/healthit/chpl/aqa/asserts/SurveillanceComplaintsAsserts.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Verify that complaint was deleted successfully.
+     * Assert that value is not found in ONC-ACB Complaint Id column in complaints view table
+     * @param acbComplaintId is value to look for in Complaint Id column
+     */
+    @Then("^the complaint with ONC-ACB Complaint ID \"([^\"]*)\" should not display in complaints view table$")
+    public void verifyComplaintIsDeleted(final String acbComplaintId) {
+        getDriver().navigate().refresh();
+        boolean isFound = true;
```
andlar: `isFound` should start as false, and switch to true when it's found. "Is found" being false when the text was found is backwards

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceComplaintsSteps.java
ashwinimore: Updated test to not select checkboxes. Also, updates are expected to these fields in UI review work

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceComplaintsSteps.java
ashwinimore: That's findElements, to find list of web elements (all acb complaint IDs) in the ONC-ACB Complaint Id column. I can't use single/specific element locator here as I am checking table data in ONC-ACB Complaint Id column row by row and then looking to click delete button based on value found in ONC-ACB Complaint Id column

src/test/java/gov/healthit/chpl/aqa/pageObjects/SurveillanceComplaintsPage.java
ashwinimore: Updated definition to use cssSelector, this was old definition that wasn't updated & wasn't used.

src/test/java/gov/healthit/chpl/aqa/pageObjects/SurveillanceComplaintsPage.java
ashwinimore: Requested locators

src/test/java/gov/healthit/chpl/aqa/asserts/SurveillanceComplaintsAsserts.java
ashwinimore: Updated test to use assertFalse

andlar (approved):

PR #163, opened 2019-08-02
test: automate announcements management

iyerashwin14:


src/test/resources/Features/OCD-2966-newAdminNavAnnouncementsManagement.feature
```diff
@@ -0,0 +1,83 @@
...
+    When I navigate to the Announcements Management page
+    And I add an Announcement
+    And I click cancel button in the Announcement section
+    Then I see "Announcements" as the page title for the Announcements management page
+    Examples:
+      |ROLE       |
+      |ROLE_ADMIN |
+      |ROLE_ONC   |
+
+Scenario Outline: Verify adding announcement is saved and recorded on Announcements page and Announcements reports page
```
iyerashwin14: This scenario is going to add Announcement on a daily basis and I could write a scenario to delete this specific announcement but there would be a dependency between two scenarios which is not correct. Can anyone suggest a better option?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/AnnouncementsSteps.java
```diff
@@ -0,0 +1,80 @@
...
+    public void addAnnouncement() {
+        JavascriptExecutor jse = (JavascriptExecutor) getDriver();
+        jse.executeScript("arguments[0].scrollIntoView()", AnnouncementsPage.addAnnouncementsButton(getDriver()));
+        AnnouncementsPage.addAnnouncementsButton(getDriver()).click();
+    }
+
+    /**
+     * Clear the "Title/Announcement-Text/StartDate/EndDate" field and enter a value.
+     * @param value to be entered
+     * @param fieldName is the name of the field in the Add Announcements section
```
andlar: fieldName is not the "name" of the field, it's the "id". It matters because the name and ID won't be the same thing if it's a multi-word thing. "title" would work, but "start-date" is not the same as "startDate"

src/test/resources/Features/OCD-2966-newAdminNavAnnouncementsManagement.feature
```diff
@@ -0,0 +1,83 @@
...
+    Given I'm logged in as "<ROLE>"
+    When I navigate to the Announcements Management page
+    And I add an Announcement
+    And I set the "title" field
```
andlar: This step should be something like `And I set the "title" field to today's date`, because a reader doesn't know what you're setting this to, and that makes it confusing, especially given that the rest of these "set the something" steps have the value that they're setting them to

andlar (changes requested):

src/test/resources/Features/OCD-2966-newAdminNavAnnouncementsManagement.feature
andlar: Delete the announcement in the same scenario. That's what was done on one of the other recent AQA PRs. It's not super great because the scenario does a lot of things, but it works, and it also verifies that "delete" also works

src/test/java/gov/healthit/chpl/aqa/pageObjects/AnnouncementsPage.java
```diff
@@ -45,5 +45,14 @@ public static WebElement savedAnnouncementInformatoin(final WebDriver driver) {
     public static WebElement announcementReportDetails(final WebDriver driver) {
         return driver.findElement(By.xpath("//*[@id=\"main-content\"]/div/ui-view/chpl-reports-announcements/div"));
     }
+
+    public static WebElement editExistingAnnouncement(final WebDriver driver) {
+        return driver.findElement(By.xpath("//tr[2]//td[6]//button[1]"));
```
ashwinimore: This looks like xpath to 2nd row and 6th column in announcement table. This won't work when there's no data in the table(like the announcement you assumed was added, did not add) or only one announcement in the table. If you are adding an announcement and then editing that specific announcement, and if it was to be in different row, you'd be editing a non intended announcement. Looping through table elements  and clicking 'Edit' button based on data in other columns (the announcement you intend to edit) is better way to go for it.

ashwinimore (changes requested):

src/test/resources/Features/OCD-2966-newAdminNavAnnouncementsManagement.feature
```diff
@@ -0,0 +1,89 @@
...
+    When I navigate to the Announcements Management page
+    And I add an Announcement
+    And I click cancel button in the Announcement section
+    Then I see "Announcements" as the page title for the Announcements management page
+    Examples:
+      |ROLE       |
+      |ROLE_ADMIN |
+      |ROLE_ONC   |
+
+Scenario Outline: Verify adding announcement is saved and recorded on Announcements page and Announcements reports page
```
andlar: This scenario also verifies that delete works, but doesn't mention that in the scenario definition

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/AnnouncementsSteps.java
```diff
@@ -0,0 +1,113 @@
...
+    public void addAnnouncement() {
+        JavascriptExecutor jse = (JavascriptExecutor) getDriver();
+        jse.executeScript("arguments[0].scrollIntoView()", AnnouncementsPage.addAnnouncementsButton(getDriver()));
+        AnnouncementsPage.addAnnouncementsButton(getDriver()).click();
+    }
+
+    /**
+     * Clear the "Title/Announcement-Text/StartDate/EndDate" field and enter a value.
+     * @param value to be entered
+     * @param fieldName is the id of the field in the Add Announcements section
```
andlar: Given that there is both a "name" and an "id" associated with form fields, it would be helpful to future developers to have the name of the variable match whatever part of the form field the variable is talking about: `fieldId`, perhaps, or, to be even closer to what it really is, maybe something like `inputId`

src/test/java/gov/healthit/chpl/aqa/pageObjects/AnnouncementsPage.java
```diff
@@ -0,0 +1,54 @@
...
+    public static WebElement annoouncementsColumnName(final WebDriver driver) {
+        return driver.findElement(By.xpath("//*[@id=\"main-content\"]/div/ui-view/chpl-announcements/div/div"));
+    }
+
+    public static WebElement addAnnouncementsButton(final WebDriver driver) {
+        return driver.findElement(By.xpath("//button[contains(text(),'Add')]"));
+    }
+
+    public static WebElement announcementField(final WebDriver driver, final String fieldName) {
+        return driver.findElement(By.xpath("//input[@id='" + fieldName + "']"));
```
andlar: Similar thing with the name of the variable

andlar (changes requested):

andlar (approved):

ashwinimore (approved):

PR #164, opened 2019-08-07
test: update pageobjects post UI updates

ashwinimore:


andlar (approved):
I thought you were going to request an ID for these elements to avoid the shuffle issue? If that won't work, or is taking too long, I'm fine with this one

iyerashwin14 (approved):

ashwinimore:
I had requested IDs, but I think request got buried in Jira comments among all other updates & issues. I'll follow up on that and update when IDs are added but till then this should fix failing tests (there aren't any pending updates on complaints UI at the moment).

PR #165, opened 2019-08-09
test: automate basic surveillance reporting feature

iyerashwin14:


src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
```diff
@@ -0,0 +1,118 @@
+@Regression
+Feature: OCD-2916 Basic Surveillance Reporting
+  This test checks whether edit, split and merge developer functionality work as expected
```
andlar: Wrong description

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
```diff
@@ -0,0 +1,118 @@
...
+
+Scenario Outline: User logs in as ROLE and verifies navigation is successful to the Surveillance Report page
```
andlar: If you're testing navigation to the page working, you should also test that it doesn't, by trying as ROLE_ATL, ROLE_CMS_STAFF, and Anonymous

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
```diff
@@ -0,0 +1,118 @@
...
+    Given I'm logged in as "<ROLE>"
+    When I navigate to the surveillance reports page
+    Then I see "CHPL Surveillance" as the page title for the surveillance report page
+    Examples:
+      |ROLE       |
+      |ROLE_ACB   |
+      |ROLE_ADMIN |
+      |ROLE_ONC   |
+
+Scenario Outline: User logs in as ROLE and verifies Surveillance reports are available for UL LLC, Drummond Group, ICSA Labs and SLI Compliance
```
andlar: Should add ROLE_ONC to this scenario, and add a Scenario for a single ACB that verifies they can only see their ACB

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
```diff
@@ -0,0 +1,118 @@
...
+    Examples:
+      |ROLE       |ACB              |ACB-Year-Quarter        |Page Title       |Subtitle                                       |Confirmation Message                                               |
+      |ROLE_ADMIN |UL LLC           |UL LLC-2019-Q1          |Quarterly Report |UL LLC Quarterly Surveillance Reporting        |Are you sure you wish to delete this Quarterly Surveillance Report?|
+      |ROLE_ACB   |Drummond Group   |Drummond Group-2020-Q1  |Quarterly Report |Drummond Group Quarterly Surveillance Reporting|Are you sure you wish to delete this Quarterly Surveillance Report?|
+      |ROLE_ADMIN |SLI Compliance   |SLI Compliance-2019-Q1  |Quarterly Report |SLI Compliance Quarterly Surveillance Reporting|Are you sure you wish to delete this Quarterly Surveillance Report?|
+      |ROLE_ACB   |ICSA Labs        |ICSA Labs-2020-Q1       |Quarterly Report |ICSA Labs Quarterly Surveillance Reporting     |Are you sure you wish to delete this Quarterly Surveillance Report?|
+
+Scenario Outline: User logs in as ROLE - initiates, edits and deletes a quarterly surveillance report for ACB
```
andlar, line 21: All of the "Then"s from the previous two scenarios should be in this one. You have three scenarios that work through the same steps, so just combine them as one. Also, there's no need to do every year/quarter in these tests; they'll start failing as soon as an ACB creates a real report. I like using the 2019 Q4 from this scenario, as those likely won't be created "for real" for a while. Just use those quarters for all of them

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceReportSteps.java
```diff
@@ -0,0 +1,105 @@
...
+        js.executeScript("arguments[0].scrollIntoView();", SurveillanceReportPage.transparencyDisclosureField(getDriver()));
+        SurveillanceReportPage.saveQuarterlyReport(getDriver()).click();
+    }
+
+    /**
+     * Edit a quarterly surveillance report.
+     * @param acbNameYearQuarter is the name of ACB, Year and Quarter for which a quarterly surveillance report will be edited
+     **/
+    @And("I edit a quarterly surveillance report for \"([^\"]*)\"$")
+    public void ieditQuarterlySurveillanceReport(final String acbNameYearQuarter) {
```
andlar: If you're using camel case, `ieditQuar...` should be `iEditQuar....`

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceReportSteps.java
```diff
@@ -0,0 +1,105 @@
...
+        SurveillanceReportPage.quarterlyReportFieldInput(getDriver(), inputId).clear();
+        SurveillanceReportPage.quarterlyReportFieldInput(getDriver(), inputId).sendKeys(value);
+        SurveillanceReportPage.quarterlyReportFieldInput(getDriver(), inputId).sendKeys(Keys.TAB);
+    }
+
+    /**
+     * Click save button to save the quarterly surveillance report.
+     **/
+    @And("I save the quarterly surveillance report$")
+    public void isaveQuarterlySurveillanceReport() {
```
andlar: Same camel case

src/test/java/gov/healthit/chpl/aqa/asserts/SurveillanceReportPageAsserts.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Assert that ACBs are expected to be displayed under available reports.
+     * @param expectedACBs are Drummond Group, ICSA Labs, SLI Compliance & UL LLC
+     */
+    @Then("^I see available reports for \"([^\"]*)\"$")
+    public void iSeeReportsAvailableFor(final String expectedACBs) {
+        String actualACBs = SurveillanceReportPage.availableYearQuarterForAcbs(getDriver(), expectedACBs).getText();
+        String[] acbTempArray = actualACBs.split("\n");
+        StringBuffer buff = new StringBuffer();
+        for (String string : acbTempArray) {
```
andlar: What is this for? You're taking a string like `Name1 \n Name 2 \n Name 3`, splitting it into an array, then joining the array again into a string like `Name1 Name 2 Name 3`, right? Then seeing if the "expectedACBs" string is in that string? What the split and join? It'd work with `\n`s in the middle of the string, right?

If anything, the scenario(s) should look for each ACB on its own, because that's what a user actually sees, not this odd "concatenated list of ACB names"

andlar (changes requested):

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
```diff
@@ -0,0 +1,118 @@
...
+    When I navigate to the surveillance reports page
+    And I expand "<ACB>"
+    Then I see "<Year-Quarters>" listed for each ACB
+    Examples:
+      |ROLE       |ACB           |Year-Quarters           |
+      |ROLE_ADMIN |UL LLC        |Year Q1 Q2 Q3 Q4        |
+      |ROLE_ONC   |SLI Compliance|Year Q1 Q2 Q3 Q4        |
+      |ROLE_ADMIN |Drummond Group|Year Q1 Q2 Q3 Q4        |
+      |ROLE_ONC   |ICSA Labs     |Year Q1 Q2 Q3 Q4        |
+
```
ashwinimore: I think, instead of this and above scenario-checking list of ACBs & quarters, running following scenario to test complete workflow would be a good regression test - like expand it to cover post confirm action as well. The scenario outline does say 'initiate quarterly report' but stops at pop up verification

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
```diff
@@ -0,0 +1,118 @@
...
+    When I navigate to the surveillance reports page
+    And I expand "<ACB>"
+    And I initate a quarterly surveillance report for "<ACB-Year-Quarter>"
+    And I confirm initiaing a quarterly surveillance report
+    Then I see "<Page Title>" for the quarterly surveillance report page
+    And "<Subtitle>" for the quarterly surveillance report 
+    And I delete the quarterly surveillance report
+    Then I see "<Confirmation Message>" on clicking delete
+    And I confirm deleting a quarterly surveillance report
+    Then I see "CHPL Surveillance" as the page title for the surveillance report page
```
ashwinimore: This should use "Available reports" instead of "CHPL Surveillance". CHPL Surveillance shows on quarterly report page as well, assert won't fail even if delete fails. Also, asset should really be on Q# button label to verify report was deleted than page title, it could be that redirect happened but delete did not succeed/complete

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
ashwinimore: agree with combining scenarios, end-to end workflow tests

ashwinimore (changes requested):

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
iyerashwin14: For ROL_ATL and ROLE_CMS_STAFF should I impersonate these users first in order to add the above test or is there any other option?

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
iyerashwin14: How can I add a Scenario for a single ACB as the AQA ACB user log in can see all the ACBs?

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
andlar: One option is to create one or more AQA users that have those roles. Another would be impersonation, which would be fine. The last option would be to not test CMS_STAFF and ATL, but just do anonymous. Our goal with AQA stuff isn't 100% coverage, at least not yet, but instead a representative sample

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
andlar: Could update the AQA user to not have all ACBs. Could impersonate a user with only one ACB. Probably should do the first, as that would make our AQA user look more like the actual ACB users in the system

src/test/resources/Features/OCD-2916-BasicSurveillanceReporting.feature
```diff
@@ -116,3 +116,7 @@ Scenario Outline: User logs in as ROLE - initiates, edits and deletes a quarterl
       |ROLE_ACB   |SLI Compliance   |SLI Compliance-2019-Q4  |Quarterly Report |
       |ROLE_ACB   |UL LLC           |UL LLC-2019-Q4          |Quarterly Report |
 
+Scenario: Surveillance reports is not available to a user who is not logged in 
+    When I navigate to the surveillance reports page
```
ashwinimore: as anonymous user

ashwinimore (changes requested):

andlar (approved):

ashwinimore (approved):

PR #171, opened 2019-08-20
test: complaints reporting text update

iyerashwin14:


src/test/java/gov/healthit/chpl/aqa/asserts/ChplSearchPageAsserts.java
```diff
@@ -194,4 +194,16 @@ public void verifyCertStatusOptionUnchecked(final String certStatusFilterOption)
...
     }
+
+    /**
+     * Assert that complaints reporting menu navigation text is correct in Surveillance drop-down.
+     * @param expectedMenuNavigationText - Complaints Reporting
+     */
+    @Then("^I see \"([^\"]*)\" as the menu navigation text for Complaints Reporting$")
+    public void iSeeComplaintsReportingMenuNavigationText(final String expectedMenuNavigationText) {
+        WebElement link = SearchPage.complaintsReporting(getDriver());
+        ((JavascriptExecutor) getDriver()).executeScript("arguments[0].scrollIntoView();", link);
```
andlar: Is this actually necessary? I don't know that the elements need to be in view to get their text, and I'm not even positive that they actually scroll into view, as I think they just scroll into the viewport, and when they do that they're often hidden behind the sticky header or footer

src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
```diff
@@ -210,4 +210,12 @@ public static WebElement loadChplID(final WebDriver driver, final String chplID)
     public static WebElement certStatusFilterOptions(final WebDriver driver) {
         return driver.findElement(By.cssSelector("#filters > div.col-sm-8 > div.btn-group.dropdown.open > ul"));
     }
+
+    public static WebElement surveillanceDropdown(final WebDriver driver) {
+        return driver.findElement(By.xpath("//button[@id='surveillance-toggle']"));
+    }
+
+    public static WebElement complaintsReporting(final WebDriver driver) {
+        return driver.findElement(By.xpath("//*[@id=\"surveillance-dropdown-menu\"]/li[4]/a"));
```
andlar: Specifying the fourth element is brittle. Would instead rather have you have the method return a list of the nav links that are in this dropdown, and then loop through them to see if the title appears

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/asserts/ChplSearchPageAsserts.java
iyerashwin14: Nope, it's not necessary in this case!

src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
iyerashwin14: Now, I'm using a different page object instead of writing a method.

src/test/java/gov/healthit/chpl/aqa/asserts/ChplSearchPageAsserts.java
andlar: Did you check the other places where this "scrollintoview" thing happens?

src/test/java/gov/healthit/chpl/aqa/pageObjects/SearchPage.java
andlar: This is now looking for any link that has the title "Complaints Reporting", right? I think that's too broad. There ought to be some xpath / selector / something that gets to `surveillance menu > has the right title`, or something like that

src/test/java/gov/healthit/chpl/aqa/asserts/ChplSearchPageAsserts.java
iyerashwin14: Yes, I did and it works for the other scenarios! For example:- I have used it while adding an Announcement

iyerashwin14: Sorry, I see what you mean and yes I removed it from the other step as well

src/test/java/gov/healthit/chpl/aqa/asserts/SurveillanceComplaintsAsserts.java
```diff
@@ -117,4 +118,13 @@ public void verifyCriteriaWasAssociatedToComplaint(final String criteria) {
         String actualValue = SurveillanceComplaintsPage.associatedCriteriaValue(getDriver()).getText();
         assertTrue(actualValue.contains(criteria), "Expect " + criteria + " to be found in " + actualValue);
     }
+
+    @Then("^I see \"([^\"]*)\" as the page title on Complaints Reporitng page$")
+    public void iSeeComplaintsReportingPageTitle(final String expectedPageTitle) {
+        WebElement link = SurveillanceComplaintsPage.complaintsReportingPageTitle(getDriver(), expectedPageTitle);
```
ashwinimore: why not use getTitle() and assert?

ashwinimore (changes requested):

andlar (approved):

src/test/java/gov/healthit/chpl/aqa/asserts/SurveillanceComplaintsAsserts.java
iyerashwin14: I'm using getText() instead of getTitle() and it's doing the same thing, please let me know if you want me to use getTitle() instead of getText()

ashwinimore: It just seemed simple to use getTitle() instead of defining page object, getting text, storing it in a string and compare. Also, i am not entirely sure why WebElement link def is needed here. Does getText() of complaintsReportingPageTitle and asset (last two lines in this method) not work? The text is at top of the page, so I was thinking it should be visible when page loads.

ashwinimore (approved):

src/test/java/gov/healthit/chpl/aqa/asserts/SurveillanceComplaintsAsserts.java
iyerashwin14: Used getTitle() and verified locally - works as expected

PR #174, opened 2019-09-10
test: scenario to add surveillance data in quarterly report

ashwinimore:


src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceReportSteps.java
```diff
@@ -111,5 +111,43 @@ public void iEditQuarterlySurveillanceReport(final String acbNameYearQuarter) {
...
+        SurveillanceReportPage.surveillancePotentialCauses(getDriver()).sendKeys(getCurrentDate());
+        SurveillanceReportPage.surveillanceadditionalCostEvaluation(getDriver()).sendKeys(getCurrentDate());
+        SurveillanceReportPage.surveillanceLimitationsEvaluation(getDriver()).sendKeys(getCurrentDate());
+        SurveillanceReportPage.surveillanceCompletedCapVerification(getDriver()).sendKeys(getCurrentDate());
+    }
+
+    @When("^I click Save Surveillance Data button$")
+    public void clickSaveSurveillanceDataButton() {
+        WebElement button =  SurveillanceReportPage.saveSurveillanceDataButton(getDriver());
+        ((JavascriptExecutor) getDriver()).executeScript("arguments[0].scrollIntoView(true);", button);
```
andlar, line 147: Is the "scrollintoview" required? I thought we had troubles with this before, as the scroll would bring the element into the frame, but it was still behind the "sticky footer".

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceReportSteps.java
```diff
@@ -111,5 +111,43 @@ public void iEditQuarterlySurveillanceReport(final String acbNameYearQuarter) {
...
+    }
+
+    @And("^I enter surveillance data in fields on form: \"([^\"]*)\" \"([^\"]*)\" \"([^\"]*)\" \"([^\"]*)\" \"([^\"]*)\" \"([^\"]*)\" \"([^\"]*)\"$")
+    public void enterSurveillanceData(final String outcome, final String processType, final String grounds,
+            final String causes, final String additionalCosts, final String limitationsEvaluation,
+            final String completedCapVerification) {
+        getDriver().manage().window().maximize();
+        SurveillanceReportPage.surveillanceOutcomeDropdown(getDriver()).sendKeys(outcome);
+        SurveillanceReportPage.surveillanceProcessTypeDropdown(getDriver()).sendKeys(processType);
+        SurveillanceReportPage.groundsForInitiatingInput(getDriver()).sendKeys(getCurrentDate());
```
andlar, line 137: It looks like most of the parameters aren't being used. Is that on purpose?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceReportSteps.java
```diff
@@ -111,5 +111,43 @@ public void iEditQuarterlySurveillanceReport(final String acbNameYearQuarter) {
     public void navigateToSurveillanceReportsPageAsAnonymousUser() {
         getDriver().get(getUrl() + "#/surveillance/reporting");
     }
-}
 
+    @And("^I click open Listings with relevant surveillance accordion$")
+    public void openListingsWithRelevantSurveillanceAccordion() {
+        WebElement button = SurveillanceReportPage.listingsWithRelevantSurveillanceAccordion(getDriver());
+        ((JavascriptExecutor) getDriver()).executeScript("arguments[0].scrollIntoView(true);", button);
```
andlar, line 118: Also the scroll. If it's not required I think it just clutters up the code

src/test/java/gov/healthit/chpl/aqa/pageObjects/SurveillanceReportPage.java
```diff
@@ -10,64 +10,128 @@
...
     }
 
-    public static WebElement initiateButton(final WebDriver driver, final String acbNameYearQuarter) {
-        return driver.findElement(By.xpath("//button[@id='initiate-" + acbNameYearQuarter + "']"));
+    public static WebElement surveillancePageTitle(final WebDriver driver) {
+        return driver.findElement(By.xpath("//h1[contains(text(),'CHPL Surveillance')]"));
+    }
+
+    public static WebElement transparencyDisclosureField(final WebDriver driver) {
+        return driver.findElement(By.xpath("//textarea[@id='transparency-disclosure']"));
```
andlar: Was there a reason not to use "By.id"?

src/test/java/gov/healthit/chpl/aqa/pageObjects/SurveillanceReportPage.java
```diff
@@ -10,64 +10,128 @@
...
+    public static WebElement editSurveillanceReport(final WebDriver driver, final String acbNameYearQuarter) {
+        return driver.findElement(By.xpath("//button[@id='act-" + acbNameYearQuarter + "']"));
+    }
+
+    public static WebElement groundsForInitiatingInput(final WebDriver driver) {
+        return driver.findElement(By.id("grounds-for-initiating"));
+    }
+
+    public static WebElement initiateButton(final WebDriver driver, final String acbNameYearQuarter) {
+        return driver.findElement(By.xpath("//button[@id='initiate-" + acbNameYearQuarter + "']"));
```
andlar: If there's an ID, is the xpath required? Same with editSurveillanceReport and editSurveillanceDataButton

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceReportSteps.java
ashwinimore: I tried not using it but kept getting element not found, with scrollintoview, the element was found, so I went with it . I can look for alternatives if it doesn't work out well.

src/test/java/gov/healthit/chpl/aqa/pageObjects/SurveillanceReportPage.java
ashwinimore: I didn't add that element definition, just rearranged what was already there, but agree that it should be ID. I'll update that

src/test/java/gov/healthit/chpl/aqa/pageObjects/SurveillanceReportPage.java
ashwinimore: Same, this one was also already there. I shall review all the page objects and update

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceReportSteps.java
ashwinimore: yes. I parameterized the test in case we want to expand the test to test with different inputs but with current testing practice to ensure that form has latest saved update, like we use on other forms, date is used as input.

athomas-git (approved):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/SurveillanceReportSteps.java
andlar: I really don't like having variables in a method that aren't used, especially when they're there "just in case" we need them later. That's a serious code smell. I'd rather you either remove the variables, or use them, but the midway thing is not desirable

andlar (approved):

PR #184, opened 2019-11-06
test: navigation to shortcuts pages is successful and displays correct page title

iyerashwin14:


athomas-git (approved):

src/test/resources/Features/OCD-3161-VerifyShortcutsPagesLoadSuccessfully.feature
```diff
@@ -0,0 +1,33 @@
...
+Scenario: Verify navigation to Inactive Certificates page is successful and displays correct page title
+    Given I am on Inactive Certificates page
+    Then I see "Inactive Certificates" as the page title for the Inactive Certificates page
+
+Scenario: Verify navigation to Corrective Action Status page is successful and displays correct page title
+    Given I am on Corrective Action Status page
+    Then I see "Products: Corrective Action Status" as the page title for the Corrective Action Status page
+
+Scenario: Verify navigation to SED Info for 2015 Products page is successful and displays correct page title
+    Given I am on SED Info for 2015 Products page
```
andlar: This step name doesn't match the scenario (nor do the others). If the scenario is testing navigation to the page, then there can't be a "Given" that you're already on the page. If you want to test navigation, then there are two tests to make; one is from a different page, using the dropdown menu, the other is using the URL directly, though I think that's not as important a test as using the navigation. These should be something more like:

> Given a user is on the Overview page
> When they navigate to the "SED Information for 2015 Edition Products" shortcut
> Then they should see "the title, or whatever"

That "When" above should be a single step that passes in the text of the link in the shortcut menu navigation, then it can be used by all of the tests

Oh, and why are there all of these blank lines at the end of the file? I know it's a minor thing, and if I didn't have other comments I wouldn't bother, but it just looks messy

src/test/java/gov/healthit/chpl/aqa/pageObjects/InactiveCertificatesPage.java
```diff
@@ -0,0 +1,18 @@
...
+import org.openqa.selenium.WebElement;
+
+/**
+ * Class InactiveCertificatesPage definition.
+ */
+public final class InactiveCertificatesPage extends BasePage {
+    private InactiveCertificatesPage() {}
+
+    public static WebElement inactiveCertificatesPageTitle(final WebDriver driver) {
+        return driver.findElement(By.xpath("/html/body/ai-compare-widget/div[2]/ui-view/div/div[1]/div/h1/ng-transclude/ai-title"));
```
andlar: Why does this locator use the `ai-title` at the end, but the SedInfoFor2015Page.java doesn't?

Also, all of the collections pages have the same structure; could they be combined into a single "CollectionsPage.java" class?

Even the steps are almost the same; is there any way to reduce that duplication? Maybe with just a CollectionsBaseSteps.java that extends from Base.java, and has that step for "navigate to a collections page", and then have an InactiveCertificatesSteps that extends from the Collections steps class if there are specific steps needed for those activities

src/test/java/gov/healthit/chpl/aqa/pageObjects/ChartsPage.java
```diff
@@ -23,5 +23,9 @@ public static WebElement chartsHealthITFeedback(final WebDriver driver) {
     public static WebElement chartsInfoText(final WebDriver driver) {
         return driver.findElement(By.xpath("//*[@id=\"main-content\"]/div[1]/div/p"));
     }
+
+    public static WebElement chartsPageTitle(final WebDriver driver, final String pageTitle) {
```
andlar: This looks like the kind of definition that might fit in the Base Page class; there's nothing in this xpath that's specific to the Charts page

src/test/java/gov/healthit/chpl/aqa/asserts/SedInfoFor2015PageAsserts.java
```diff
@@ -0,0 +1,19 @@
...
+import gov.healthit.chpl.aqa.stepDefinitions.Base;
+
+/**
+ * Class SedInfoFor2015PageAsserts definition.
+ */
+public class SedInfoFor2015PageAsserts extends Base {
+
+    @Then("^I see \"([^\"]*)\" as the page title for SED Info for 2015 Products page$")
+    public void iSeePageTitleForSedInfoFor2015Page(final String expectedPageTitle) {
+        String actualPageTitle = SedInfoFor2015Page.sedInfoFor2015PageTitle(getDriver()).getText();
```
andlar: These might collapse into a single "CollectionsAsserts" for now, if the page title is made a single element in the CollectionsPage class. We can extend the asserts later when we have tests that are different between different collections pages

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/CollectionsPagesSteps.java
```diff
@@ -47,5 +48,11 @@ public void downloadSEDDetailsFile(final String fileName) throws FileNotFoundExc
         ((JavascriptExecutor) getDriver()).executeScript("arguments[0].click();", link);
         super.checkIfFileIsDownloaded(fileName);
     }
-}
 
+    @Given("^I navigate to \"([^\"]*)\" page$")
+    public void userLoadsCollectionsPage(final String pageName) {
+        getDriver().get(getUrl() + "/#/collections/" + pageName);
```
andlar: This action does not match the description of the step. A user "navigating" to a page does not do so by entering the URL into their bar, they do it by clicking on the Shortcuts dropdown, then on the name of the page they want to go to. Or by starting on the search page and clicking the big grey box for the collection they want.

andlar (changes requested):
I really like all of the changes that you made. There's just one more thing that I'd like to see done differently

andlar (approved):

PR #186, opened 2019-11-18
test: verify api returns correct status code

iyerashwin14:


src/test/resources/Features/verifyApiReturnsCorrectStatusCode.feature
```diff
@@ -0,0 +1,161 @@
+@Regression
```
iyerashwin14: All the scenarios can be combined to a single one but it could be a bit confusing while trying to add more endpoints in near future. That's one of the reasons I have separated the scenarios based on the controllers.

palakpatel311: Very new to this but why can't we have feature file per API? it will be easier to maintain. whenever we get new change to that API, we just have to touch that feature file.

src/test/resources/Features/api-test/announcementController.feature
```diff
@@ -0,0 +1,8 @@
+@Regression
+Feature: Verify announcement-controller endpoints return correct status code
+
+Scenario Outline: Verify announcement-controller endpoints return correct status codes for a non-logged in user
+    Then the "<API>" endpoint returns correct "<Status Code>"
+    Examples:
+      |API                                 |Status Code|
+      |/announcements                      |200        |
```
andlar: This is the only one that doesn't have a newline at the end of the file

andlar (changes requested):

src/test/resources/Features/api-test/announcementController.feature
iyerashwin14: Done!

andlar (approved):

palakpatel311 (approved):

PR #187, opened 2019-11-21
Refactoring chromedriver and properties

palakpatel311:
Couple of changes:
1. Added Chromedriver in project it self
2. Added properties file to avoid passing VM arguments from eclipse for each feature
3. Corrected login test which was failing

### DIFF
--- a/prop.properties
+++ b/prop.properties
@@ -1,0 +1,8 @@
+url = https://chpl.ahrqdev.org/
+apikey = 12909a978483dfb8ecd0596c98ae9094
+roleAdminUsername= aqa-admin
+roleAdminPassword= A good long password123
+roleAcbUsername= aqa-acb
+roleAcbPassword= A good long password123
+roleOncUsername= aqa-onc
+roleOncPassword=A good long password123
--- a/src/test/java/gov/healthit/chpl/aqa/pageObjects/LoginPage.java
+++ b/src/test/java/gov/healthit/chpl/aqa/pageObjects/LoginPage.java
@@ -5,6 +5,10 @@ import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 
 /**
+ * 		11/21/2019	Palak Patel	: Removed and updated welcome Text webelement
+ */	
+
+/**
  * Class LoginPage definition.
  */
 public final class LoginPage extends BasePage {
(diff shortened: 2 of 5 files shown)


prop.properties
```diff
@@ -0,0 +1,8 @@
+url = https://chpl.ahrqdev.org/
```
andlar, line 1: We can't set the URL permanently in a property file because we sometimes want to run tests against DEV, sometimes against STG, sometimes against PROD, and sometimes against local environments. That's why that isn't a property

prop.properties
```diff
@@ -0,0 +1,8 @@
...
+apikey = 12909a978483dfb8ecd0596c98ae9094
+roleAdminUsername= aqa-admin
+roleAdminPassword= A good long password123
```
andlar, line 4: Should never commit passwords in git, and also should not push them to github. This is a public repository and once a password is pushed it's public. I've disabled all of the AQA users, and we'll need to change their passwords to different ones before they can be re-enabled

src/test/java/gov/healthit/chpl/aqa/pageObjects/LoginPage.java
```diff
@@ -4,6 +4,10 @@
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 
+/**
+ * 		11/21/2019	Palak Patel	: Removed and updated welcome Text webelement
```
andlar, line 8: These kinds of changes are better seen using git and git history. Comments like these are only a little helpful when first made, and in 3 to 6 months they'll be so out of date they won't be of any use at all

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -47,20 +57,9 @@ public void openBrowser() {
...
-                // If temp directory creation failed, create new directory in target folder
-                // user.dir - User working directory, make new directories in user's working directory
-                File file = new File("target", "download-files");
-                file.mkdirs();
-                tempDirectory = System.getProperty("user.dir") + File.separator + "target" + File.separator + "download-files";
-            }
-            downloadPath = tempDirectory;
-        }
+
+        downloadPath = System.getProperty("user.dir") + File.separator;
```
andlar, line 61: Why get rid of the temporary directory functionality?

andlar (changes requested):
We shouldn't save the chromedriver in git, because that driver only works on Windows machines, and all of our Jenkins and Bamboo machines are linux ones, which can't use that chromedriver at all

palakpatel311:
I ll just have to then close this pull request and i will update corrected login test as part of different pull request.

PR #190, opened 2019-12-05
Scheduled jobs feature

iyerashwin14:


src/test/java/gov/healthit/chpl/aqa/pageObjects/ScheduledJobsPage.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Class ScheduledJobsPage definition.
+ */
+public final class ScheduledJobsPage extends BasePage {
+    private ScheduledJobsPage() {
+    }
+
+    public static WebElement typesOfSheduledJobs(final WebDriver driver) {
+        WebElement webelement  = driver.findElement(By.xpath("//table[@class='table table-condensed']//tbody"));
```
andlar, line 15: I think there are two tables on this page that will meet this xpath, but you might not see the first one unless there are actually some jobs that are schedueld

src/test/java/gov/healthit/chpl/aqa/pageObjects/ScheduledJobsPage.java
```diff
@@ -0,0 +1,30 @@
...
+        return webelement;
+    }
+
+    public static List<WebElement> getRowElements(final WebElement row) {
```
andlar: Both of these "getRowElements" and "getColElements" look like they should be in the BasePage file; there's nothing on either of these specific to the scheduledJobs page

src/test/java/gov/healthit/chpl/aqa/asserts/ScheduledJobsPageAsserts.java
```diff
@@ -0,0 +1,60 @@
...
+     * Assert types of jobs listed in the scheduled jobs page.
+     * @param jobNameValue is the expected job name
+     * @param descriptionValue is the expected description
+     * @param oncAcbSpecificValue is the expected OncAcbSpecific
+     * @param jobTypeValue is the expected JobType
+     */
+    @Then("^I see types of jobs with \"([^\"]*)\" , \"([^\"]*)\" , \"([^\"]*)\" and \"([^\"]*)\"$")
+    public void scheduledJobsList(final String jobNameValue, final String descriptionValue, final String oncAcbSpecificValue, final String jobTypeValue) {
+        boolean jobFound = false;
+        WebElement element = ScheduledJobsPage.typesOfSheduledJobs(getDriver());
```
andlar: The variable name "element" tells me nothing about what's in this variable. My guess is that it's a list, because that's what the PageObject call looks like it would return

src/test/java/gov/healthit/chpl/aqa/asserts/ScheduledJobsPageAsserts.java
```diff
@@ -0,0 +1,60 @@
...
+        List<List<String>> actualValue = tableElements(element);
```
andlar: What is an "actualValue"?

src/test/java/gov/healthit/chpl/aqa/asserts/ScheduledJobsPageAsserts.java
```diff
@@ -0,0 +1,60 @@
...
+        for (List<String> rows : actualValue) {
```
andlar: "rows" suggests that you're looping through a bunch of tables and doing something on each table's set of rows at once. I think this variable should be "row"

src/test/java/gov/healthit/chpl/aqa/asserts/ScheduledJobsPageAsserts.java
```diff
@@ -0,0 +1,60 @@
...
+                jobFound = true;
+            }
+            if (jobFound) {
+                break;
+            }
+        }
+        assertTrue(jobFound, "Job not found in the types of jobs list");
+    }
+
+    private List<List<String>> tableElements(final WebElement row) {
```
andlar: It helps if the method name says what it's doing. Something like "getColumnsFromTableRow", though I'm not actually sure if that's what this is doing

andlar (changes requested):

src/test/resources/Features/OCD-3159-scheduledJobs.feature
```diff
@@ -0,0 +1,40 @@
+@Regression
+Feature: OCD-3159 - Types of Jobs listed in Scheduled Jobs page
+  This test checks whether the correct jobs are listed in scheduled jobs page based on user role
```
ashwinimore, line 3: the scenario that is added might be good as a regression test to check all jobs are available/displayed to specific user (currently only ADMIN; ONC access may be added as a part of this ticket or later and then this test will need an update),  but that is not what the ticket is for. For this ticket, relevant test is to check scheduled system job triggers display on screen (after it's scheduled). If the goal is to check access based on role then adding a test that ROLE ACB does not see any of the systemJobs is a good idea (test that simply looks for systemJobs text not found on page).

src/test/java/gov/healthit/chpl/aqa/pageObjects/BasePage.java
```diff
@@ -59,4 +61,14 @@ public static WebElement headerNav(final WebDriver driver) {
     public static WebElement pageTitle(final WebDriver driver, final String pageTitle) {
         return driver.findElement(By.xpath("//h1[contains(text(),'" + pageTitle + "')]"));
     }
+
+    public static List<WebElement> getRowElements(final WebElement row) {
```
andlar: If you were to describe this method in English, it would be something like:

> Given a table
> Return all of the rows in that table

Right? If that's the case, then the "input" WebElement shouldn't be named "row", but instead be named "table", and the output element should definitely not be "roWebElements" (also note the missing w in "row") but instead just be "rows".

Similar comments for "getColElements"

src/test/java/gov/healthit/chpl/aqa/asserts/ScheduledJobsPageAsserts.java
```diff
@@ -0,0 +1,66 @@
...
+                jobFound = true;
+            }
+            if (jobFound) {
+                break;
+            }
+        }
+        assertTrue(jobFound, "Job not found in the types of jobs list");
+    }
+
+    private List<List<String>> getRowListFromTable(final WebElement row) {
```
andlar: This method name implies you're getting a list of rows, but you're not, you're getting a nested list of elements. I could see "getTableElements" as a reasonable name, as well as "getElementsWithinRows", but when I read this name in the method that calls it, and see that the variable it's assigned to is "tableRows", I definitely did not expect it to already have parsed the rows into lists of column elements

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/pageObjects/BasePage.java
```diff
@@ -59,4 +61,14 @@ public static WebElement headerNav(final WebDriver driver) {
     public static WebElement pageTitle(final WebDriver driver, final String pageTitle) {
         return driver.findElement(By.xpath("//h1[contains(text(),'" + pageTitle + "')]"));
     }
+
+    public static List<WebElement> getRowElements(final WebElement table) {
+        List<WebElement> rows = table.findElements(By.tagName("tr"));
+        return rows;
+    }
+
+    public static List<WebElement> getColElements(final WebElement table) {
```
andlar: This method is called with `List<WebElement> colElts = ScheduledJobsPage.getColElements(col);` but the variable in the parameter is `table`

I don't think "col" is correct, but that's for a different comment. Here the problem is the "table"

src/test/java/gov/healthit/chpl/aqa/asserts/ScheduledJobsPageAsserts.java
```diff
@@ -0,0 +1,66 @@
...
+                break;
+            }
+        }
+        assertTrue(jobFound, "Job not found in the types of jobs list");
+    }
+
+    private List<List<String>> getElementsWithinRows(final WebElement row) {
+        List<List<String>> rowValues = new ArrayList<List<String>>();
+        List<WebElement> roWebElements = ScheduledJobsPage.getRowElements(row);
+        for (WebElement col : roWebElements) {
```
andlar: The previous line gets rows, right? This line loops through those rows, so why is the variable called "col"?

src/test/java/gov/healthit/chpl/aqa/pageObjects/BasePage.java
iyerashwin14: I'm not sure whether the changes that I pushed now looks good. If not, can you please suggest the names?

andlar (approved):

ashwinimore (approved):

PR #193, opened 2019-12-12
Adding complaints controller tests

palakpatel311:
1. Restructuring UI vs API test as per QA meeting
2. Adding crud E2E test to complaints controller
NOTE: I know some of the method can be moved to Base but i don't have enough knowledge about other controllers.

Let me know which methods we can move to Base.


src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -38,7 +45,7 @@
     /**
      * Launch ChromeDriver.
      */
-    @Before
+    @Before("@Regression")
```
andlar: Does this mean that UI tests can only be run as part of the @Regression suite of tests?

palakpatel311: Yes we separated tags for UI and API test.
UI test will be tagged as @Regression and API will be tagged as @RegressionAPI

andlar: Right, but what if I don't want to run Regression tests at all? What if I just want to run a specific feature file, or a specific test tagged @WIP? Will these Before and After steps not run if I don't also include the @Regression tag?

palakpatel311: Yes you need to add those tags to this before and after.

andlar: If that's the case then the README.md file will need to be updated to indicate that, and the various samples of how to run "a feature file" or "a tag" or some combination of both or whatever will need to be figured out too

palakpatel311: I have updated it. There were minor changes.

README.md
```diff
@@ -59,7 +60,7 @@ Options can be used to tweak the functionality of the tests. They're used with t
 
 To run only a single feature file, run the command
 
-`mvn test -Dcucumber.options="src/test/resources/Features/filename.feature"`
+`mvn test -Dcucumber.options="src/test/resources/Features/UI OR API/filename.feature"`
```
andlar, line 63: If this is a UI feature, this won't work without having the @Regression tag applied as well, right?

Would it make more sense to use a tag like @UI on those Before and After methods, and leave the @Regression tag alone? That way one could run all Regression tests with @Regression @UI, just the API ones with @Regression, and if you want to run a feature file that is related to UI tests you'd only need to add the @UI tag?

What I don't want to do is try to run two or three UI tests marked as @WIP, but be required to add the @Regression tag to do so, which would then mean I'd be running all of the Regression tests as well

palakpatel311: the whole point of having separate tags is because i dint want to invoke chrome browser every time i run API test. so we have to separate UI and API tags. we cant keep same tag for both.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
palakpatel311: I have pushed work around for this issue. now before and after will ignore tags for RegressionAPI and will run for all other tags. Readme is also updated to reflect that

andlar (approved):

ashwinimore (approved):

PR #195, opened 2019-12-18
OCD-3116 validate compliant status based on closed date

athomas-git:


src/test/java/gov/healthit/chpl/aqa/asserts/SurveillanceComplaintsAsserts.java
```diff
@@ -122,5 +123,26 @@ public void verifyCriteriaWasAssociatedToComplaint(final String criteria) {
...
+        boolean isFound = false;
+
+        WebElement table = SurveillanceComplaintsPage.complaintsTable(getDriver());
+        List<WebElement> rows = table.findElements(By.xpath("tbody/tr"));
 
+        for (WebElement row : rows) {
+            List<WebElement> cols = row.findElements(By.xpath("td"));
+            if (cols.get(ACBCOMPLAINTID_COL).getText().equals(acbComplaintId)) {
+                isFound = true;
+                break;
```
andlar: The "Then" here says that the complaint with ID <something> AND status <something> is found, but if the code gets to the point, where it finds the ID, it will then return true and break out of the loop, but it has not checked that the status matches <something>

andlar (changes requested):

andlar (approved):

src/test/resources/Features/surveillanceComplaints.feature
```diff
@@ -72,3 +73,18 @@ Scenario: Associate a criteria to a complaint
     And I click save button
     And I click Edit complaint for ONC-ACB Complaint ID "ONCACB005"
     Then the criteria "170.315 (a)(1)" should be associated to the complaint
+    
+Scenario: Validate compliant status based on closed date
```
ashwinimore: 'complaint'- I don't want to hold off the PR for typos but it'd be nice to fix those if you can in this one

src/test/resources/Features/surveillanceComplaints.feature
```diff
@@ -5,6 +5,7 @@ Feature: Verify complaints can be added
          OCD-2913-Associate a complaint with a listing
          OCD-2914-Associate a complaint with criteria
          OCD-2915-Associate a complaint with surveillance
+         OCD-3116-Validate compliant status based on closed date
```
ashwinimore: Complaint

ashwinimore (approved):

PR #199, opened 2019-12-31
test: update complaints tests per new requirement

ashwinimore:


src/test/resources/Features/UI/surveillanceComplaints.feature
```diff
@@ -68,7 +68,7 @@ Scenario: Associate a criteria to a complaint
     Given I'm logged in as "ROLE_ACB"
     And I am on complaints page
     And I click Add New Complaint button
-    And I enter complaint data in fields: "SLI Compliance" "07/12/2019" "07/12/2019" "ONCACB005" "ONC001" "Developer" "TBD"
+    And I enter complaint data in fields: "SLI Compliance" "07/12/2019" "" "ONCACB005" "ONC001" "Developer" "TBD"
```
andlar: I think it would make more sense to remove the "close date" field entirely, as opposed to setting it to blank. The only reason to not do so is if the method that handles this step also handles edit, and so needs that field, but I don't know if that's the case

src/test/resources/Features/UI/surveillanceComplaints.feature
```diff
@@ -35,18 +35,18 @@ Scenario: Delete a complaint
...
+    Then the complaint with ONC-ACB Complaint ID "D$DC" should not display in complaints view table
     
 Scenario: Associate a listing to a complaint
     Given I'm logged in as "ROLE_ACB"
     And I am on complaints page
     And I click Add New Complaint button
-    And I enter complaint data in fields: "SLI Compliance" "07/12/2019" "07/12/2019" "ONCACB002" "ONC001" "Developer" "TBD"
-    When I associate listing "15.05.05.2760.ISAD.01.00.0.180412" to the complaint
+    And I enter complaint data in fields: "SLI Compliance" "07/12/2019" "" "ONCACB002" "ONC001" "Developer" "TBD"
+    When I associate listing "15.05.05.2760.ISAD.01.00.0.18041" to the complaint
```
andlar: I don't think this CHPL ID is valid; it ends with only 5 characters

andlar (changes requested):

src/test/resources/Features/UI/surveillanceComplaints.feature
ashwinimore: I realized that right after creating PR and updated tests to remove closed date from add complaint step

andlar (approved):

palakpatel311 (approved):

PR #203, opened 2020-01-08
Crossbrowser

palakpatel311:
1. Running test on firefox and edge on local
2. Running test on firefox on jenkins
3. Updated selenium webdriver version to support geckodriver


ashwinimore (approved):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
-    public static void takeScreenshot() throws Exception {
-        String hash = String.valueOf(new Date().getTime());
-        takeScreenshot(hash);
-    }
-    
-
-}
+		 */
+
+		if(System.getenv("browser")!=null && !System.getenv("browser").isEmpty()){
```
andlar: There are a lot of code style issues that checkstyle should be warning about. I see a bunch of tabs while the team has decided spaces is what we're using. Also, there should be a space after the `if` and before the `(`, as well as around `!=` and other operators

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
+			browser = System.getenv("browser");
+			System.setProperty("webdriver.gecko.driver", System.getProperty("PathtoGeckodriver") + File.separator + "geckodriver");
```
andlar: Shouldn't the code check to see if the browser is gecko before setting a property on the gecko webdriver?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
-        /* To run chrome with the developer tools window automatically opened re-able these lines.
-         *
+	private static File dir;
+	private static EventFiringWebDriver driver;
+	public static Properties config = new Properties();
+	private static final int DELAY = 30;
+	private static String screenshotPath;
+	public static FileInputStream fis;
+	private static String downloadPath = System.getProperty("downloadPath");
+	public static String browser;
```
andlar: Why is this public?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
+		if(System.getenv("browser")!=null && !System.getenv("browser").isEmpty()){
+			browser = System.getenv("browser");
+			System.setProperty("webdriver.gecko.driver", System.getProperty("PathtoGeckodriver") + File.separator + "geckodriver");
+
+		}else{
+			fis = new FileInputStream(
+					System.getProperty("user.dir") + File.separator+"src"
+							+File.separator+"test"+File.separator+"resources"
+							+File.separator+"Config.properties");
+			config.load(fis);
```
andlar: Why is this config file loaded only if there's no browser defined?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
+			browser = config.getProperty("browser");
+
+			System.setProperty("webdriver.gecko.driver", System.getProperty("PathtoGeckodriver") + File.separator + "geckodriver.exe");
```
andlar: This will only work on Windows, right? Is this setting up code that defaults to windows if there's no config file defined? If that's the case, then the config file that's in this PR means that this code will never run

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
+
+		}
+		config.setProperty("browser", browser);
+		if (StringUtils.isEmpty(downloadPath)) {
+			String tempDirectory;
+			try {
+				tempDirectory = Files.createTempDirectory("download-files").toString();
+				// Print the path to the newly created directory
```
andlar: Nothing prints anything

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
+	public static FileInputStream fis;
+	private static String downloadPath = System.getProperty("downloadPath");
+	public static String browser;
+	/**
+	 * Launch ChromeDriver.
+	 * @throws IOException 
+	 */
+	@Before("~@RegressionAPI")
+	public void openBrowser() throws IOException {
+		/* To run chrome with the developer tools window automatically opened re-able these lines.
```
andlar: These lines probably need to be moved to the chrome section of the method; they won't work for firefox, right?

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
+			options.setExperimentalOption("prefs", chromePrefs);
+			chromePrefs.put("safebrowsing.enabled", "true");
+			options.addArguments("--safebrowsing-disable-download-protection");
+			options.addArguments("disable-popup-blocking");
+			options.setCapability(CapabilityType.ACCEPT_SSL_CERTS, true);
+			screenshotPath = System.getProperty("user.dir") + File.separator + "test-output";
+			driver = new EventFiringWebDriver(new ChromeDriver(options));
+		}
+
+		else if(config.getProperty("browser").equals("firefox")) {
```
andlar: This line should be on line 104, as `} else if (config...`, and checkstyle should have warned about this

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
+			driver = new EventFiringWebDriver(new FirefoxDriver());
```
andlar: Are there no properties that need to be enabled / disabled for Firefox? For all of the ones that chrome has that seems odd

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
+		}
+
+		else if(config.getProperty("browser").equals("edge")) {
+			driver =new EventFiringWebDriver( new EdgeDriver());
```
andlar: Are there no properties that need to be enabled / disabled for Edge? For all of the ones that chrome has that seems odd

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -1,147 +1,166 @@
...
+		File scrFile = ((TakesScreenshot) getDriver()).getScreenshotAs(OutputType.FILE);
+		File outFile = new File(screenshotPath + File.separator + "failed-test-" + hash.replaceAll("\\W+", "_") + ".png");
+		FileHandler.copy(scrFile, outFile);
+	}
+
+	/**
+	 * Take a screenshot. Generates hash based on current time.
+	 * @throws Exception if screenshot cannot be created
+	 */
+	public static void takeScreenshot() throws Exception {
```
andlar: I don't know why all of these lines are showing as changed. Is it because of the change from spaces to tabs? That's making it hard to tell what actually changed

src/test/resources/config.properties
```diff
@@ -0,0 +1 @@
+browser=chrome
```
andlar: By putting this config file in git, it makes it hard for different users to use different configs. Anyone who tried to use a different browser would have to commit a change to the config. This kind of thing should be in either an ignored config file, or read via a command line argument

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
palakpatel311: This else part is handling while running locally and if part is handling input from jenkins job(this is set up on my fork-jenkins).
So when running locallly, we need to load config file otherwise no need.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
palakpatel311: Above comment will answer this as well.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
palakpatel311: I did not see a need to add any options now. Once we run whole automation suite with firefox once, i will go through and see if there is any need to add any firefoxoptions.

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
palakpatel311: Same as firefox with this one

src/test/resources/config.properties
palakpatel311: Added to gitignore

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
palakpatel311: nothing changed here. its checkstyle

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
```diff
@@ -38,27 +31,40 @@
...
+            }
+        } else {
+            fis = new FileInputStream(
+                    System.getProperty("user.dir") + File.separator + "src"
+                            + File.separator + "test" + File.separator + "resources"
+                            + File.separator + "Config.properties");
+            config.load(fis);
+            browser = config.getProperty("browser");
+            if (browser.equalsIgnoreCase("firefox")) {
+                System.setProperty("webdriver.gecko.driver", System.getProperty("PathtoGeckodriver") + File.separator + "geckodriver.exe");
```
andlar: If we have to get the path to the gecko driver from the 'getproperty' call anyway, why not just put the full driver in that value, and include the actual file name? The way it's currently set up, we're assuming someone is running locally on Windows, but there's no reason that needs to be the case. Could move this entire statement outside of the if/else thing and leave it with 
```suggestion
                System.setProperty("webdriver.gecko.driver", System.getProperty("PathtoGeckodriver"));
```

andlar: Also, no reason not to set the property even if we're not using firefox; it won't hurt to set it (unless it fails if null), but it simplifies things if we do it automatically

src/test/resources/config.properties
```diff
@@ -0,0 +1 @@
+browser=firefox
```
andlar: If this file is ignored, then it shouldn't be in the PR

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/Hooks.java
palakpatel311: so path to gecko driver is only needed for jenkins so i have updated hooks.java now

andlar (approved):

PR #207, opened 2020-01-21
Fixing automation tests

palakpatel311:


andlar (approved):
I don't really  like hardcoding the IDs in the page objects for the surveillance complaints page, but if they work in the current set of tests I think it's okay. When/if we need to use those page objects to select different things we'll need to update those PO elements, but for now it's okay

Oh, do appreciate the commit messages; it's good to see each commit with a message that explains something about that commit

ashwinimore (approved):

PR #208, opened 2020-01-22
test: add tests to verify ACBs can't add/edit surveillance associated to 2014 criteria

ashwinimore:


src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ListingUploadConfirmSteps.java
```diff
@@ -420,11 +358,8 @@ public void inspectSurveillanceDetails(final String chplId) {
 
     /**
      * Confirm uploaded surveillance activity.
-     * 
-     * @param survChplId
-     *            is chpl id of listing for surveillance activity
-     * @throws Exception
-     *             if there is an exception
+     * @param survChplId is chpl id of listing for surveillance activity
```
andlar, line 360: One of the checkstyle rules we've discarded is having the javadocs. Unless you think this is useful information, I'm on board with just removing them, especially for super small methods like these

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ListingUploadConfirmSteps.java
```diff
@@ -235,13 +192,11 @@ public void testSedTestingEndDateDisplayedInSedDetails(final String sedEndDate)
...
      * Load listing details to verify listing was uploaded successfully.
-     *
-     * @param ed
-     *            - edition digits in CHPL ID
+     * @param edition - edition digits in CHPL ID
      */
     @Then("^I see that listing was uploaded successfully to CHPL "
             + "and listing details load as expected for uploaded \"([^\"]*)\" listing$")
-    public void testVerifyUploadWasSuccessful(final String ed) {
+    public void testVerifyUploadWasSuccessful(final String edition) {
```
andlar: The variable name changed, but nothing else in the method did. That implies the variable isn't being used, and should be removed

andlar (changes requested):

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ListingUploadConfirmSteps.java
ashwinimore: sounds good, will update with other checkstyle fixes in next PR

src/test/java/gov/healthit/chpl/aqa/stepDefinitions/ListingUploadConfirmSteps.java
```diff
@@ -235,13 +192,11 @@ public void testSedTestingEndDateDisplayedInSedDetails(final String sedEndDate)
...
      * Load listing details to verify listing was uploaded successfully.
-     *
-     * @param ed
-     *            - edition digits in CHPL ID
+     * @param edition - edition digits in CHPL ID
      */
     @Then("^I see that listing was uploaded successfully to CHPL "
             + "and listing details load as expected for uploaded \"([^\"]*)\" listing$")
-    public void testVerifyUploadWasSuccessful(final String ed) {
+    public void testVerifyUploadWasSuccessful() {
```
andlar, line 198: The step passes in a variable. If the method doesn't need the variable then the step should be updated to not pass it in

andlar (approved):

palakpatel311 (approved):

PR #212, opened 2020-02-11
test : Adding announcement controller test

palakpatel311:


src/test/java/gov/healthit/chpl/aqa/api/stepDefinitions/ComplaintsTest.java
```diff
@@ -31,38 +31,33 @@ public void setHeaders(String authorization) {
...
 
-    @When("^I send DELETE request to resource \"([^\"]*)\"$")
-    public void sendDeleteRequest(String resource) {
-        response = request.pathParams("complaintId", ComplaintsAsserts.postComplaintId).when().delete(resource);
-    }
+//    @When("^I send DELETE request to resource \"([^\"]*)\"$")
+//    public void sendDeleteRequest(String resource) {
+//        response = request.pathParams("complaintId", ComplaintsAsserts.postComplaintId).when().delete(resource);
+//    }
 }
```
ashwinimore, line 61: why comment and keep when a new step & step definition is added to replace this step?

src/test/resources/Features/API/announcementController.feature
```diff
@@ -1,9 +1,29 @@
...
+Feature: Announcement-controller endpoints
+  Verify CRUD operations on announcement-controller endpoints
+  
+Background: Setting base URL
+    Given I set base URL
 
+  Scenario: Verify adding, updating and deleting a announcement
+    Given I set Headers with API key, "ROLE_ADMIN" authorization for announcements 
+    And I set "POST" announcements request body
+    When I send "POST" request to annoucement resource "rest/announcements"
```
ashwinimore: typo on announcement here and several places below

ashwinimore (approved):

src/test/java/gov/healthit/chpl/aqa/api/stepDefinitions/Base.java
```diff
@@ -14,6 +16,8 @@
     private static String username;
     private static String password;
     private static final int HTTP_GOOD_RESPONSE = 200;
+    private static RequestSpecification request;
+    public static Response response;
```
andlar: This should probably be private, and have getter/setter methods

src/test/java/gov/healthit/chpl/aqa/api/stepDefinitions/AnnoucementsTest.java
```diff
@@ -0,0 +1,39 @@
...
+import cucumber.api.java.en.Given;
+import cucumber.api.java.en.When;
+import gov.healthit.chpl.aqa.api.asserts.AnnoucementsAsserts;
+import gov.healthit.chpl.aqa.api.payLoad.AnnouncementsPayload;
+import io.restassured.response.Response;
+import io.restassured.specification.RequestSpecification;
+
+public class AnnoucementsTest {
+
+    public static Response response;
```
andlar: Should also be private

src/test/java/gov/healthit/chpl/aqa/api/asserts/AnnoucementsAsserts.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import org.testng.Assert;
+import cucumber.api.java.en.Then;
+import gov.healthit.chpl.aqa.api.stepDefinitions.AnnoucementsTest;
+import io.restassured.path.json.JsonPath;
+import io.restassured.response.Response;
+
+public class AnnoucementsAsserts {
+    private Response response;
+    public static int postAnnoucementId;
```
andlar: Think this should be private as well

andlar (commented):
Do all of these tests work as written? I know one thing that I saw as challenging was figuring out a way to pass information between steps in a Feature file; it looks like you're doing that here with a static class; is that generally accepted as the right way of doing that?

palakpatel311:
@andlar Yes these test works. But i see confusion. I need to use same response/request and pass it through steps which became tough to do it this way. I could write feature file more consolidated way to avoid this. let me know if you have any other suggestion

andlar (approved):

